PR #8, opened 2015-05-28
Changes for eap 7

mkrajcov:
refactor of test case classes


heiko-braun:
Are you developing this against HAL master?

mkrajcov:
I am developing this against eap 7 with hal version 2.6.9.Final

heiko-braun:
You should be using 2.7.0-SNAPSHOT, which reflects the master branch (intended to be used with EAP7)

rhatlapa:
I have tried running the tests locally and I am getting some failures [1].

There is less failures than when running tests with version of this testsuite without changes done by this PR.

[1]

```
Failed tests: 
  SocketBindingsTestCase.changeInboundMulticastPort:115 Attribute value is different in model. expected:<[45700]> but was:<[undefined]>
  SocketBindingsTestCase.changeInboundPort:105 Attribute value is different in model. expected:<[43150]> but was:<[undefined]>
Tests in error: 
  SocketBindingsTestCase.before:66 » Timeout Timed out after 10 seconds waiting ...
  ModClusterTestCase.flushPackets:236 » ElementNotVisible Element is not current...
```

src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/TestConnectionTestCase.java
```diff
@@ -80,12 +79,14 @@ public void invalidDatasource() {
         testConnection(dsNameInvalid, false);
     }
 
+    @Ignore
```
rhatlapa, line 82: Use Ingores with messages explaining why specific test is ignored.

src/test/java/org/jboss/hal/testsuite/test/util/ConfigAreaUtils.java
```diff
@@ -0,0 +1,165 @@
...
+import org.jboss.hal.testsuite.page.ConfigPage;
+import org.jboss.hal.testsuite.util.ResourceVerifier;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author mkrajcov <mkrajcov@redhat.com>
+ */
+public class ConfigAreaUtils {
```
rhatlapa: I would prefer name of class which doesn't contain Utils as this is not standard utility class. E.g. something like `ConfigAreaChecker` or something else

rhatlapa:
I ran the tests with the new commits (only in standalone mode) and they are passing now, thanks.

The PR looks generally ok and ready to merge, could you just rename the class `ConfigAreaUtils` to different name or do you have any objections to using non Utils name for it?

mkrajcov:
renamed, thanks for your suggestions

rhatlapa:
Thank you, merging.

PR #10, opened 2015-07-08
New navigation abstraction

mkrajcov:
Using ConfigurationPage it is possible to easily navigate through Configuration in hal 2.8+


heiko-braun:
very cool. tnx

PR #13, opened 2015-07-14
Update of current test cases

mkrajcov:
Update of navigation in test cases


rhatlapa:
I have design note: I believe it would be better to use the same interface/class for standalone and domain whenever possible, the page should only differ at initialization. Most tests are tested against default setting, which is defined once. And thus it can be done at single place. This is true at least for the tests in Shared category. For tests specific for domain, the domain config class could be initialized and used directly making use of its additional domain specific functionality.

rhatlapa:
Per discussion with Michal we shall rework the standalone vs domain design afterwards.

PR #14, opened 2015-07-15
Updated Datasources tests

honza-kasik:
Updated DomainTestConnectionTestCase to work with current navigation
- DomainConfigurationPage as entry point
- updated to proper navigation to before()
- updated to proper navigation in testValidWithSameNameInOtherGroup()

Updated WizardWindow to use dynamic waiting instead of fixed one (was causing test not to pass in some cases).

Updated DataSourcesOperations to use only one proper command instead of two, because there was an error in CLI when the first one was entered without `--xa-datasource-properties=`

```
[Server:server-three] 09:21:43,208 ERROR [org.jboss.as.controller.management-operation] (ServerService Thread Pool -- 72) WFLYCTL0013: Operation ("add") failed - address: ([
[Server:server-three]     ("subsystem" => "datasources"),
[Server:server-three]     ("xa-data-source" => "bla")
[Server:server-three] ]) - failure description: "WFLYJCA0069: At least one xa-datasource-property is required for an xa-datasource"
```

Updated TestConnectionTestCase to proper navigation in before() and removed `@Ignore`s

### DIFF
--- a/src/main/java/org/jboss/hal/testsuite/page/config/ConfigurationPage.java
+++ b/src/main/java/org/jboss/hal/testsuite/page/config/ConfigurationPage.java
@@ -2,6 +2,8 @@ package org.jboss.hal.testsuite.page.config;
 
 import org.jboss.arquillian.graphene.Graphene;
 import org.jboss.arquillian.graphene.findby.ByJQuery;
+import org.jboss.arquillian.graphene.wait.AttributeBuilder;
+import org.jboss.arquillian.graphene.wait.ElementBuilder;
 import org.jboss.hal.testsuite.cli.Library;
 import org.jboss.hal.testsuite.fragment.ConfigAreaFragment;
 import org.jboss.hal.testsuite.fragment.shared.modal.ConfirmationWindow;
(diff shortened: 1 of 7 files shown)


rhatlapa:
It was created against HAL 2.8.0, Navigation changed in HAL 2.8.2 => don't merge now, should be updated accordingly.

heiko-braun:
@rhatlapa which change in 2.8.2 is causing the problem? maybe we can improve the test stability by adding more reliable test ID's to the DOM structure?

rhatlapa:
One of the changes is different way to choose whether you are configuring datasource vs xa-datasource. Adding ID in this case would not help.

honza-kasik:
Datasources test cases were updated. They work now with no problems.

heiko-braun:
@rhatlapa ah, right. I was thinking you refer to general changes

src/main/java/org/jboss/hal/testsuite/page/config/ConfigurationPage.java
```diff
@@ -2,6 +2,8 @@
 
 import org.jboss.arquillian.graphene.Graphene;
 import org.jboss.arquillian.graphene.findby.ByJQuery;
+import org.jboss.arquillian.graphene.wait.AttributeBuilder;
+import org.jboss.arquillian.graphene.wait.ElementBuilder;
```
rhatlapa: These imports are not used

src/main/java/org/jboss/hal/testsuite/page/config/ConfigurationPage.java
```diff
@@ -21,7 +23,7 @@
     public ConfigurationPage select(String label) {
         String cellClass = PropUtils.get("table.cell.class");
         String cellSelectedClass = PropUtils.get("table.cell.selected.class");
-        By selector = ByJQuery.selector("." + cellClass + ":contains('" + label + "')");
+        By selector = By.ByXPath.xpath("//td[contains(@class,'" + cellClass + "') and descendant::div[@class='navigation-column-item' and text()='" + label + "']]");
```
rhatlapa, line 24: Wouldn't it be possible to use rather jquery even for equals? I did quick google search and found this: https://api.jquery.com/attribute-equals-selector/

rhatlapa, line 24: Per discussion with @honza-kasik, we need text of subelement and there wasn't found another simple way using jquery to do equals for subelement. Thus using XPath for handling this case.

PR #15, opened 2015-07-15
Fixing deployments test cases

cypo721:
Fixed both managed and unmanaged deployment test cases for both standalone and domian mode.
Fixed for jbosss7.0.0 DR6 version.


src/main/java/org/jboss/hal/testsuite/page/config/ConfigurationPage.java
```diff
@@ -40,11 +40,19 @@ public void remove() {
         }
     }
 
+    public void unassign() {
```
mkrajcov: Please move this to DeploymentPage as it is specific for deployments only

src/test/java/org/jboss/hal/testsuite/test/runtime/deployments/DomainManagedDeploymentsTestCase.java
```diff
@@ -57,7 +62,8 @@ public static void cleanUp() {
 
     @Before
     public void before() {
-        browser.navigate().refresh();
+        Graphene.goTo(HomePage.class);
```
mkrajcov: You can use Console.withBrowser(browser).refreshAndNavigate method to minimize redundancy in &#64; Before

zebrik:
Can you please remove unused imports? 
I see some in
org.jboss.hal.testsuite.page.runtime.DomainDeploymentPage
org.jboss.hal.testsuite.page.runtime.DeploymentPage
as well as in all testcases in org.jboss.hal.testsuite.test.runtime.deployments package

PR #22, opened 2015-08-05
Updated  files because of graphene testsuite

cypo721:
Updated closeFile method, added names in FinderName and updated label.properies.

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/finder/FinderNames.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/finder/FinderNames.java
@@ -33,12 +33,14 @@ public interface FinderNames {
     String DEPLOYMENT = "Deployment";
     String HOST = "Host";
     String HOSTS = "Hosts";
+    String LOG_FILES="Log Files";
     String MONITOR = "Monitor";
     String PROFILE = "Profile";
     String PROFILES = "Profiles";
     String SERVER = "Server";
     String SERVER_GROUP = "Server Group";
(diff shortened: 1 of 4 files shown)


src/main/java/org/jboss/hal/testsuite/page/RuntimePage.java
```diff
@@ -1,13 +1,14 @@
...
 import org.jboss.hal.testsuite.fragment.runtime.ServerContextPicker;
+import org.jboss.hal.testsuite.page.config.ConfigurationPage;
 import org.jboss.hal.testsuite.util.Console;
 import org.openqa.selenium.WebElement;
 
 /**
  * Created by mvelas on 31.3.2014.
  */
-public abstract class RuntimePage extends BasePage {
+public abstract class RuntimePage extends ConfigurationPage {
```
zebrik: Please revoke this change. As we discussed face to face RuntimePage shouldn't extended Configuration change but rather methods from ConfigurationPage which can be shared should be pulled up to BasePage.

common/src/main/java/org/jboss/hal/testsuite/page/RuntimePage.java
```diff
@@ -1,6 +1,7 @@
 package org.jboss.hal.testsuite.page;
 
 import org.jboss.hal.testsuite.fragment.runtime.ServerContextPicker;
+import org.jboss.hal.testsuite.page.config.ConfigurationPage;
```
zebrik: Please remove the unused import.

PR #23, opened 2015-08-05
Updated JGroup tests

honza-kasik:
Updated JGroups test for current testsuite.
There was an executor setting in settings in DR6 but it is not present in DR7. I just added Ignore annotation for now - if it is not coming back, I can remove it completely.


zebrik:
I will close this PR, fix the API incompatibilities (moved category package) and do a new MR

PR #24, opened 2015-08-07
HAL-786 Make HAL testsuite modular

zebrik:
Motivation: ability to clearly separate configuration for e.g. RBAC related tests


zebrik:
@rhatlapa please review
@mkrajcov @honza-kasik  @cypo721 @hpehl @heiko-braun FYI

basic/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+    <version>0.0.2-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hal-testsuite-basic</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.jboss.as</groupId>
+      <artifactId>hal-testsuite-common</artifactId>
+      <version>${project.version}</version>
```
rhatlapa: Putting the version to dependency management in parent might be better.

rbac/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+    <version>0.0.2-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hal-testsuite-rbac</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.jboss.as</groupId>
+      <artifactId>hal-testsuite-common</artifactId>
+      <version>${project.version}</version>
```
rhatlapa: The same as above

zebrik:
I agree that putting the version to dependency management in parent would be better.
Force pushed to pom.xml basic/pom.xml rbac/pom.xml

zebrik:
Just have updated README yet - info about modules.

rhatlapa:
Looks good to me, merging.

PR #28, opened 2015-08-11
Added webMetrics and webServiceEndpoints test cases

cypo721:
Moved metrics test cases from graphene.
Updated Location of metrics page and getting metrics area.


common/src/main/java/org/jboss/hal/testsuite/page/runtime/WebMetricsPage.java
```diff
@@ -7,9 +7,9 @@
 /**
  * Created by mkrajcov on 4/10/15.
  */
-@Location("#web-metrics")
+@Location("#http-metrics")
```
zebrik: IMO this `@Location` annotation can be removed. Anyway neither "#web-metrics" nor "#http-metrics" should be used.

cypo721:
Location deleted, commit squashed.

PR #32, opened 2015-08-13
Updated finish method in wizard to work with "Done" labeled button

honza-kasik:
Updated finish method to work with "Done" labeled button.


rhatlapa:
Looks good to me, merging.

PR #34, opened 2015-08-19
Added option to FinderNavigation to select row by its exact content

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/finder/FinderNavigation.java
```diff
@@ -95,7 +106,25 @@ public Row selectRow() {
...
+            throw new IllegalStateException("No address for selecting a row given.");
+        }
+        return Graphene.createPageFragment(Row.class, row);
+    }
+
     private WebElement[] navigate() {
+        return navigate(false);
+    }
+
+    /**
```
zebrik: Please remove this duplicate javadoc. It's not necessary in private method. Everything else seems OK for me.

honza-kasik: Removed

PR #36, opened 2015-08-19
Hal 797

zebrik:
Add tests for hierarchical composition of profiles.

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/profiles/ProfileOperations.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/profiles/ProfileOperations.java
@@ -1,0 +1,99 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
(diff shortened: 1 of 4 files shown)


zebrik:
@rhatlapa  can you please review this?

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/profiles/ProfileOperations.java
```diff
@@ -0,0 +1,99 @@
...
+import org.jboss.hal.testsuite.dmr.AddressTemplate;
+import org.jboss.hal.testsuite.dmr.DefaultContext;
+import org.jboss.hal.testsuite.dmr.Dispatcher;
+import org.jboss.hal.testsuite.dmr.Operation;
+import org.jboss.hal.testsuite.dmr.StatementContext;
+
+/**
+ * Created by pjelinek on Aug 18, 2015
+ */
+public class ProfileOperations {
```
rhatlapa, line 40: Short javadocs for class as is this one would be great.

As this is class which is sharable for everyone needing to test parts concerning profiles.

common/src/main/java/org/jboss/hal/testsuite/finder/FinderNavigation.java
```diff
@@ -50,19 +50,36 @@
         }
     }
 
+    public interface Hook{
```
rhatlapa: we should better enforce good code style, here is missing space before `{`

zebrik:
I have force pushed requested changes.

rhatlapa:
Thanks, merging

PR #38, opened 2015-08-21
For large domain we are more interested about column presence than vi…

zebrik:
…sibility.


zebrik:
@rhatlapa please review.

rhatlapa:
Thank you, merging

PR #42, opened 2015-08-27
Fixed WebMetricsTestCase

honza-kasik:


basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -66,7 +67,11 @@ public void before() throws IOException,TimeoutException,InterruptedException {
         statementContext = new DefaultContext();
         ResourceAddress addressStats = ADDRESS_TEMPLATE_STATISTICS.resolve(statementContext);
         dispatcher.execute(new Operation.Builder(ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION, addressStats).param("name", "statistics-enabled").param("value", "true").build());
-        cliClient.reload();
+
+        if (cliClient.reloadRequired()) {
+            cliClient.reload();
+            Library.letsSleep(5000);//navigation starts before all servers in domain are reloaded => hangs after clicking on server-one row
```
rhatlapa: This should be done without need to hardcoding sleeps. To minimize impact for now, I suggest to move the sleep only for case when running the test in domain. 

Note: There is plan to update hal testsuite to use creaper (https://github.com/Creaper-Project/creaper), when there is done its first release to community. Creaper already has support for waiting also for servers in domain to be running and not only the host after reload.

honza-kasik: Sleep replaced with timeout.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -93,6 +96,18 @@ public void before() throws IOException, TimeoutException, InterruptedException
         Console.withBrowser(browser).waitUntilLoaded();
     }
 
+    private void waitUntilRunning() throws TimeoutException {
+        String command = "/host=master/server=server-one:read-attribute(name=server-state)";
```
rhatlapa: you didn't define anywhere which are the servers, hardcoding isn't good thing to do. Having list of server for which to wait to start would make more sense.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -93,6 +96,18 @@ public void before() throws IOException, TimeoutException, InterruptedException
...
+        int timeout = 60000;
+        while (!cliClient.executeForResult(command).contains("running")) {
+            if (timeout <= 0) {
+                throw new TimeoutException();
```
rhatlapa: It is better to add message to the thrown exception explaining what happened.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
honza-kasik: Should I add a static String variable with server name then? (In this case, only server-one is important for navigation.) Or should I add more general method to e.g. BasePage? (Which will become obsolete with [creaper release](https://github.com/Creaper-Project/creaper))

rhatlapa: I suggest renaming the method to reflect that it is waiting for server-one and mark in comment that it should be handled more generally later on

honza-kasik:
Updated

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -93,6 +96,23 @@ public void before() throws IOException, TimeoutException, InterruptedException
         Console.withBrowser(browser).waitUntilLoaded();
     }
 
+    /**
+     * Waits until server-one is in running state. This should be replaced by more general method in future.
+     * @throws TimeoutException
+     */
+    private void waitUntilServerOneRunning() throws TimeoutException {
```
rhatlapa: Please move this method to `org.jboss.hal.testsuite.cli.DomainManager` and I suggest making it parametrized, where the parameter would be `serverName`.

honza-kasik:
Updated

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
```diff
@@ -472,4 +503,113 @@ public boolean waitUntilAvailable(String host, long timeout) {
...
+
+    /**
+     * Filters out all servers which are not running
+     * @param servers list of servers from which will be filtered non-running servers
+     * @return List of all running servers
+     */
+    public List<String> filterAllRunningServers(List<String> servers) {
+        List<String> running = new LinkedList<>();
+        for (String server : servers) {
+            Library.letsSleep(100);
```
rhatlapa: why break between requests? I don't see any reason for it.

common/src/main/java/org/jboss/hal/testsuite/cli/ServerManager.java
```diff
@@ -82,5 +82,19 @@ public String getVersion() {
...
+     * @param timeout max time to wait for server to start, in milliseconds
+     */
+    public abstract void reloadAndWaitUntilRunning(long timeout);
+
+    /**
+     * Reloads server(s), which require reload and waits until server(s) is(are) running again
+     *
+     * @param timeout max time to wait for server to start, in milliseconds
+     */
+    public abstract void reloadIfRequiredAndWaitUntilRunning(long timeout);
```
rhatlapa: Don't see any reason for this method as CliClient does this by default, see `org.jboss.hal.testsuite.cli.CliClient#reload()`

common/src/main/java/org/jboss/hal/testsuite/cli/StandaloneManager.java
```diff
@@ -36,10 +39,21 @@
 
     private static Logger log = LoggerFactory.getLogger(StandaloneManager.class);
 
+    private Dispatcher dispatcher = new Dispatcher();
+
     public StandaloneManager(CliClient cliClient) {
         super(cliClient);
     }
 
+    public StandaloneManager(CliClient client, Dispatcher dispatcher) {
```
rhatlapa: Combining usage of CliClient and Dispatcher is dangerous, don't do that if possible. Also ServerManager classes are used from CliClient.

common/src/main/java/org/jboss/hal/testsuite/cli/ServerManager.java
honza-kasik: It only waits for host's availability which could cause problems.

honza-kasik:
Updated

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
```diff
@@ -472,4 +475,105 @@ public boolean waitUntilAvailable(String host, long timeout) {
...
+    @Override
+    public void reloadIfRequiredAndWaitUntilRunning(long timeout) {
+        if (isReloadRequired(listRunningServers(getDefaultHost()))) {
+            reloadAndWaitUntilRunning(timeout);
+        }
+    }
+
+    @Override
+    public void reloadAndWaitUntilRunning(long timeout) {
+        List<String> servers = filterAllRunningServers(listRunningServers(getDefaultHost()));
```
rhatlapa, line 496: This seems as duplicite call, as the `listRunningServers` should return running servers, why do you need filter them again to filter only those running?

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
```diff
@@ -472,4 +475,105 @@ public boolean waitUntilAvailable(String host, long timeout) {
         }
         return isRunning;
     }
+
+    @Override
+    public void reloadIfRequiredAndWaitUntilRunning(long timeout) {
```
rhatlapa, line 484: I suggest changing name, in order to not be confusing with general `waitUntilAvailable` which checks only Host.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -52,7 +52,7 @@
     private CliClient cliClient = CliClientFactory.getClient();
     private FinderNavigation navigation;
     private DefaultContext statementContext;
-    private Dispatcher dispatcher;
+    private static Dispatcher dispatcher;
```
rhatlapa: Don't see any reason, why it should be suddenly `static`. Having it static doesn't make sense as you are initializing it again and again in `@Before` method

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
honza-kasik, line 496: Actually, I think that the `listRunningServers` method does not provide list of servers in running state but only list of servers which are assigned under given host. I did not write this method. Should I do refactoring then?

heiko-braun, line 496: I think @honza-kasik is right. The method should be renamed or changed to only return running servers. Changing it however may have side effects on other tests.

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
honza-kasik, line 484: To be honest I did not come with anything better and I think that after adding `Servers` word in method name would make it too long. Is not the javadoc sufficient in this case? Or what name would you suggest?

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
rhatlapa, line 496: You are right, it is just confusing name. I would leave the refactoring to different PR if decided to be done.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
honza-kasik: Sorry, my mistake. I did not see it after I tried something with `@BeforeClass`.

common/src/main/java/org/jboss/hal/testsuite/cli/DomainManager.java
rhatlapa, line 484: the `Servers` word would be fine, still I am ok also with leaving this as is.

rhatlapa:
You can squash the commits a little bit, otherwise +1 for merging.

honza-kasik:
Commits squashed.

common/src/main/java/org/jboss/hal/testsuite/cli/StandaloneManager.java
```diff
@@ -128,4 +127,41 @@ public boolean waitUntilAvailable(long timeout) {
...
+        }
+    }
+
+    private boolean isReloadRequired() {
+        return cliClient.reloadRequired();
+    }
+
+    public void reloadServer() {
+        cliClient.reload(true);
+    }
```
rhatlapa: One more thing, I overlooked, the implementation in `StandaloneManager` just mimics CliClient.reload methods. as such they are not really needed.

E.g. `waitUntilRunning` is more or less the same as `waitUntilAvailable`

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -70,9 +70,9 @@ public void before() throws IOException, TimeoutException, InterruptedException
         statementContext = new DefaultContext();
         ResourceAddress addressStats = ADDRESS_TEMPLATE_STATISTICS.resolve(statementContext);
         dispatcher.execute(new Operation.Builder(ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION, addressStats).param("name", "statistics-enabled").param("value", "true").build());
-        cliClient.reload();
 
         if (ConfigUtils.isDomain()) {
+            ServerManager.getInstance(cliClient).reloadAndWaitUntilRunning(60000);
```
rhatlapa: In this case initializing directly `DomainManager` would make more sense, as the `reloadAndWaitUntilRunning` was added to standalone even though not needed as `CliClient.reload` automatically waits for server to start and in standalone mode there is no confusion with waiting only for host vs all servers as in domain.

rhatlapa:
Sorry for not noticing earlier, when fixed, you can also squash the commits if you find it desirable.

common/src/main/java/org/jboss/hal/testsuite/cli/StandaloneManager.java
honza-kasik: I overlooked it too. `waitUntilRunning` is gone.

honza-kasik:
Now it is using instance of `DomainManager`. Should I squash the commits little bit more, or it is OK?

common/src/main/java/org/jboss/hal/testsuite/cli/StandaloneManager.java
rhatlapa: All changes in the `StandaloneManager` doesn't add anything and as such create only duplicate code. As written before, the reload `cliClient.reload` operations be default wait for server to become available, additional checking for the same doesn't make sense.

rhatlapa:
Thank you, merging.

common/src/main/java/org/jboss/hal/testsuite/cli/StandaloneManager.java
honza-kasik: All changes in `StandaloneManager` were removed.

PR #49, opened 2015-09-07
Added test for checking name in Assing content of domain deployments …

cypo721:
…in DomainDeployment Testcas.

Updated DomainDeploymentPage.


zebrik:
Thank you.

PR #55, opened 2015-09-17
Updated FinderNavigation without refreshing page.

cypo721:


common/src/main/java/org/jboss/hal/testsuite/util/Console.java
```diff
@@ -287,6 +287,12 @@ public void pageDown() {
         return this;
     }
 
+    public <T extends BasePage> Console navigate(Class<T> clazz) {;
```
zebrik: please remove this nomore needed method

zebrik:
thx

PR #57, opened 2015-09-18
Rbac tests created

cypo721:
Finished all rbac test cases for standalone  and domain mode.


common/src/main/java/org/jboss/hal/testsuite/util/Console.java
```diff
@@ -287,6 +287,16 @@ public void pageDown() {
         return this;
     }
 
+    public <T extends BasePage> Console waitForFirstNavigationPanel(Class<T> clazz) {
```
zebrik, line 290: Could this method call `refreshAndNavigate` and then wait? To avoid code duplication and increase readability.

common/src/main/java/org/jboss/hal/testsuite/finder/FinderNavigation.java
```diff
@@ -202,6 +204,10 @@ public FinderNavigation resetNavigation() {
         return this;
     }
 
+    public void clearNavigation() {
```
zebrik, line 213: Please add javadoc for this method to emphasize the difference compared to `resetNavigation` method and when to use this method.

zebrik:
Apart from two details (see comments above) LGTM.

common/src/main/java/org/jboss/hal/testsuite/util/Console.java
cypo721, line 290: Method updated.

common/src/main/java/org/jboss/hal/testsuite/finder/FinderNavigation.java
cypo721, line 213: Javadoc added.

zebrik:
Thank you!

PR #61, opened 2015-10-02
Fixed jca policy testcases.

cypo721:
Saving of property for incrementer/decrementer is OK, but there is still problem with Incrementer/Decrementer class in Resource Adapters subsystem.


zebrik:
Thank you.

PR #64, opened 2015-10-05
Updated most of infinispan tests

honza-kasik:
Although some fails still persists...


zebrik:
LGTM apart of unused import in HibernateCachePage ... will fix it in separate PR

rhatlapa:
unused imports should be fixed directly not in separate commits. One of the good git practices is that code should be compilable and runnable with each git commit. I suggest in such cases do manual merge, which would incorporate such fix or wait for the creator of the PR to fix it.

PR #67, opened 2015-10-08
Fixed resource adapters test case

honza-kasik:
Changed names of input fields to right values and removed editing of deprecated attributes.


zebrik:
Thank you.

PR #69, opened 2015-10-09
Finally fixed infinispan tests. Fixed standalone mode and cache selec…

honza-kasik:
…tion problem. There were also added few test for cache container settings. Some test still failing due to the HAL-873.

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/infinispan/InvalidationCacheTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/infinispan/InvalidationCacheTestCase.java
@@ -1,7 +1,7 @@
 package org.jboss.hal.testsuite.test.configuration.infinispan;
 
 import org.jboss.arquillian.junit.Arquillian;
-import org.jboss.hal.testsuite.category.Standalone;
+import org.jboss.hal.testsuite.category.Shared;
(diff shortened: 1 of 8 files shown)


honza-kasik:
Please, do not merge yet, I noticed another mistake...

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/infinispan/InvalidationCacheTestCase.java
```diff
@@ -12,7 +12,7 @@
  * @author mkrajcov <mkrajcov@redhat.com>
  */
 @RunWith(Arquillian.class)
-@Category(Standalone.class)
+@Category(Shared.class)
```
honza-kasik, line 15: Added Shared, although the navigation for domain mode persisted from Standalone, so the navigation for Domain mode wont work...

honza-kasik:
Updated: fixed navigation for cache containers test case.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/infinispan/InvalidationCacheTestCase.java
honza-kasik, line 15: Fixed

zebrik:
Thank you!

PR #73, opened 2015-10-14
Created test cases for handlers which needs file.

cypo721:
Created File handler, Periodic size handler and Size handler test cases.
Updated Logging page.


zebrik:
Thank you!

PR #75, opened 2015-10-15
Fixed and extended transaction subsystem tests

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/page/config/TransactionsPage.java
```diff
@@ -1,12 +1,32 @@
...
 @Location("#transactions")
-public class TransactionsPage extends ConfigPage {
+public class TransactionsPage extends ConfigPage implements Navigatable {
+
+    public void navigate() {
+        FinderNavigation navigation;
+        if (ConfigUtils.isDomain()) {
+            navigation = new FinderNavigation(browser, DomainConfigEntryPoint.class)
+                    .addAddress(FinderNames.CONFIGURATION, FinderNames.PROFILES)
+                    .addAddress(FinderNames.PROFILE, "full");
```
zebrik: Please use `ConfigUtils.getDefaultProfile()` instead of hardcoded "full" profile name.

zebrik:
Thank you.

PR #84, opened 2015-10-26
Fixed syslog handler testcase.

cypo721:
Updated creation of syslog handler.


common/src/main/java/org/jboss/hal/testsuite/page/config/LoggingPage.java
```diff
@@ -85,11 +85,17 @@ public void addAsyncHandler(String name, String queueLen) {
     public void addFileHandler(String name) {
         clickButton("Add");
         getWindowFragment().getEditor().text("name", name);
-        getWindowFragment().clickButton("Next");
+        getWindowFragment().clickButton("Save");
```
zebrik: This button label should not be changed to "Save".
BTW Wizard would have done the job better.

cypo721:
Fixed,

zebrik:
Thank you.

PR #91, opened 2015-11-05
Fixed CacheContainerTestCase for domain, removed ignored testcases fr…

honza-kasik:
…om infinispan and fixed navigation


zebrik:
Thank you.

PR #93, opened 2015-11-05
Fixed resource adapter test case.

cypo721:


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/connector/ResourceAdaptersTestCase.java
```diff
@@ -115,10 +115,9 @@ public void createProperties() {
         ConfigPropertiesFragment fragment = area.switchToProperty();
         ConfigPropertyWizard wizard = fragment.addProperty();
         Console.withBrowser(browser).waitUntilLoaded();
-
+        wizard.getEditor().text("name", PROPERTY_KEY);
```
rhatlapa, line 118: This is sudden inconsistency in property wizard dialog? If this is supposed to be different type of wizard, we should create abstraction for it, if not, we should make sure that there is consistent approach.

cypo721:
Label id = 'name' is used only in this Resource Adapters subsystem. In other subsystems is used label id = 'key' which is used in property wizard.

zebrik:
This inconsistency has been reported as https://issues.jboss.org/browse/HAL-954

zebrik:
Thank you.

PR #102, opened 2015-11-06
Fixed ServerOperations test case.

cypo721:
Changed type of exception, because navigation is wainting until element is visible, so there is timeout expection.


zebrik:
Thank you.

PR #104, opened 2015-11-11
Updated verifyAttribute methods in verifier.

cypo721:
Updated methods to asking for specified time in cycle if response is returning expected value.

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/dmr/ResourceVerifier.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/dmr/ResourceVerifier.java
@@ -24,6 +24,8 @@ package org.jboss.hal.testsuite.dmr;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.dmr.ModelNode;
 
+import java.util.Arrays;
+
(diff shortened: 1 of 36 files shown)


common/src/main/java/org/jboss/hal/testsuite/dmr/ResourceVerifier.java
```diff
@@ -119,16 +139,29 @@ public void verifyAttribute(ResourceAddress address, String attribute, String ex
...
-                .param(ModelDescriptionConstants.NAME, attribute)
-                .withTimeout(timeout).build();
+    private Operation readAttributeOperation(ResourceAddress address, String attribute) {
+            return new Operation.Builder(READ_ATTRIBUTE_OPERATION, address)
+                    .param(ModelDescriptionConstants.NAME, attribute).build();
+    }
+
+    private DmrResponse getResponseAndWait(ResourceAddress address, String attribute) {
+        DmrResponse response = dispatcher.execute(readAttributeOperation(address, attribute));
+        Library.letsSleep(50);
```
zebrik, line 163: Please switch order of lines 163 and 164 (move `sleep` before `dispatcher.execute`) to avoid waiting for 50ms longer than is necessary by avoiding last non-essential loop run.

cypo721:
Updated.

zebrik:
Thank you.

PR #106, opened 2015-11-19
Added identifiers to randomly generated strings

honza-kasik:
All randomly generated strings should have assigned "id" which helps in stabilization debugging. If you find any randomly generated string without "identifier", please report.


zebrik:
Thank you.

PR #107, opened 2015-11-19
Fixed socket binding dependency problem

honza-kasik:


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/ProcessIDTestCase.java
```diff
@@ -51,11 +51,6 @@ public void before() {
         page.getConfig().switchTo("Process ID");
     }
 
-    @AfterClass
```
zebrik, line 54: Please remove now unused import - org.junit.AfterClass as well.

honza-kasik, line 54: I removed the unused import, Thank you for noticing!

zebrik:
Thank you.

PR #108, opened 2015-11-19
Added AddSocketBinding creaper command

honza-kasik:


zebrik:
Please resolve conflicts.

honza-kasik:
Conflicts were resolved.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/undertow/UndertowOperations.java
```diff
@@ -194,9 +199,15 @@ public void removeBufferPool(String name) {
...
     public String createSocketBinding() {
         String name = "UndertowSocketBinding_" + RandomStringUtils.randomAlphanumeric(6);
-        ResourceAddress address = socketBindingAddressTemplate.resolve(context, name);
-        Map<String, String> params = ImmutableMap.of("port", String.valueOf(ThreadLocalRandom.current().nextInt(1000, 9999)));
-        executeAddAction(address, params);
+        try (OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient()) {
+            client.apply(new AddSocketBinding.Builder(name)
+                    .port(ThreadLocalRandom.current().nextInt(1000, 9999))
+                    .build());
+        } catch (CommandFailedException e) {
```
zebrik: I prefer rethrow instead of continuing to run the test. BTW you probably wanted to use `printStackTrace` instead of `getStackTrace`.

rhatlapa: Thrown exception means it failed to create the port => the socketBinding wasn't created => test should be aborted (fail).

honza-kasik: You are right, I was not aware of this, fixed... Thanks!

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/undertow/UndertowOperations.java
```diff
@@ -192,11 +197,13 @@ public void removeBufferPool(String name) {
...
 
-    public String createSocketBinding() {
+    public String createSocketBinding() throws CommandFailedException, IOException {
         String name = "UndertowSocketBinding_" + RandomStringUtils.randomAlphanumeric(6);
-        ResourceAddress address = socketBindingAddressTemplate.resolve(context, name);
-        Map<String, String> params = ImmutableMap.of("port", String.valueOf(ThreadLocalRandom.current().nextInt(1000, 9999)));
-        executeAddAction(address, params);
+        try (OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient()) {
+            client.apply(new AddSocketBinding.Builder(name)
+                    .port(ThreadLocalRandom.current().nextInt(1000, 9999))
```
rhatlapa: Ports lower than 1024 require root privilages on Linux. The random should be set to use ports available for all users.

common/src/main/java/org/jboss/hal/testsuite/creaper/command/AddSocketBinding.java
```diff
@@ -0,0 +1,156 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * @author Jan Kasik <jkasik@redhat.com>
+ *         Created on 19.11.15.
+ */
+public class AddSocketBinding implements OnlineCommand {
```
rhatlapa, line 22: The class itself should have javadoc describing its intent.

BTW: this command is nicely done and I believe is good match for having it directly in creaper.
=> you should think about creating also merge request to https://github.com/wildfly-extras/creaper.

common/src/main/java/org/jboss/hal/testsuite/creaper/command/AddSocketBinding.java
```diff
@@ -0,0 +1,156 @@
...
+        Operations ops = new Operations(ctx.client);
+
+        String socketBindingGroup = ctx.client.options().isDomain ? "full-sockets" : "standard-sockets";
+        Address socketBindingAddress = Address.of("socket-binding-group", socketBindingGroup)
+                .and("socket-binding", socketBindingName);
+
+        if (replaceExisting) {
+            try {
+                ops.removeIfExists(socketBindingAddress);
+            } catch (OperationException e) {
```
zebrik, line 64: Why is this wrapped to `IOException`? You can wrap it to another `OperationException`. @honza-kasik , @rhatlapa WDYT?

zebrik:
LGTM. One last question is about wrapping OperationException into IOException see inline comment.

common/src/main/java/org/jboss/hal/testsuite/creaper/command/AddSocketBinding.java
rhatlapa, line 64: @zebrik good point. I believe you can even throw OperationException. If you would like to wrap it, I am for CommandFailedException

honza-kasik, line 64: @zebrik , @rhatlapa Thanks, I wrapped it as CommandFailedException instead, i think I chose IOException because it was already present because of `ops.add...` but this gives more sense...

I still do not have overview over all creaper classes.

common/src/main/java/org/jboss/hal/testsuite/creaper/command/AddSocketBinding.java
zebrik, line 22: +1 for creating merge request to Creaper itself  as well.

zebrik:
Thank you!

PR #109, opened 2015-11-20
Adding Creaper ResourceVerifier

zebrik:


zebrik:
@rhatlapa please review.

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -0,0 +1,113 @@
...
+    public ResourceVerifier forResource(Address address) {
+        this.resourceAddress = address;
+        return this;
+    }
+
+    /**
+     * Verifies if previously set resource exists in model.
+     */
+    public ResourceVerifier verifyResource(boolean expected) throws IOException, OperationException {
+        addressIsSetCheck();
```
rhatlapa: As you are always expecting address to be set, why it is not a mandatory part of constructor?

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    public ResourceVerifier forResource(Address address) {
+        this.resourceAddress = address;
+        return this;
+    }
+
+    /**
+     * Verifies if previously set resource exists in model.
+     */
+    public ResourceVerifier verifyResource(boolean expected) throws IOException, OperationException {
```
rhatlapa: this method name is not clear, it would be better to use something like `verifyResourceExistence` to make clear what is being verified.

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * @throws AssertionError if previously provided resource exists and has attribute with attributeName
+     * with value equal to notExpectedValue.
+     */
+    public ResourceVerifier verifyAttributeNotEqual(String attributeName, String notExpectedValue) throws IOException {
+        addressIsSetCheck();
+        ModelNodeResult actualResult = ops.readAttribute(resourceAddress, attributeName);
+        if (actualResult.isSuccess() && actualResult.hasDefinedValue() && actualResult.stringValue().equals(notExpectedValue)) {
```
rhatlapa: Why not using assertFalse?

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -0,0 +1,113 @@
...
+            throw new AssertionError(attributeName + " should not have value " + notExpectedValue);
+        }
+        return this;
+    }
```
rhatlapa, line 114: Not all resources in model are Strings, wouldn't it make sense to add methods for List, Boolean, .... or make the comparison compare ModelNodes?

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
zebrik: I felt this way as more clear but ok lets change it to more usual way.

zebrik:
Updated. Please review again.

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -0,0 +1,124 @@
...
+import org.wildfly.extras.creaper.core.online.ModelNodeResult;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+/**
+ * Created by pjelinek on Nov 16, 2015
+ */
+public class ResourceVerifier {
```
rhatlapa, line 39: Please add javadoc

rhatlapa:
Thanks, merging

PR #112, opened 2015-11-26
Added Apache Mina library dependency to obtain an available port

honza-kasik:
Depends on pull request #108

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/undertow/UndertowOperations.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/undertow/UndertowOperations.java
@@ -2,6 +2,7 @@ package org.jboss.hal.testsuite.test.configuration.undertow;
 
 import com.google.common.collect.ImmutableMap;
 import org.apache.commons.lang.RandomStringUtils;
+import org.apache.mina.util.AvailablePortFinder;
 import org.jboss.hal.testsuite.cli.CliClientFactory;
 import org.jboss.hal.testsuite.cli.DomainManager;
 import org.jboss.hal.testsuite.creaper.ManagementClientProvider;
@@ -22,7 +23,6 @@ import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
-import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * @author Jan Kasik <jkasik@redhat.com>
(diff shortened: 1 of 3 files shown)


zebrik:
@honza-kasik please resolve conflicts .. these commits are not based upon https://github.com/hal/testsuite/pull/108 commits

honza-kasik:
Thanks, resolved.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/undertow/UndertowOperations.java
```diff
@@ -201,7 +201,7 @@ public String createSocketBinding() throws CommandFailedException, IOException {
         String name = "UndertowSocketBinding_" + RandomStringUtils.randomAlphanumeric(6);
         try (OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient()) {
             client.apply(new AddSocketBinding.Builder(name)
-                    .port(ThreadLocalRandom.current().nextInt(10000, 19999))
+                    .port(AvailablePortFinder.getNextAvailable())
```
zebrik: It seems to give ephemeral port numbers since it uses new java.net.ServerSocket(0) so should be OK but not guaranteed. It might be good idea to log selected port number. WDYT?

zebrik: May be the obtained port number can be checked to be greater than 1024 and possibly select another one. Or do you think higher port number is guaranteed?

honza-kasik: We can use `AvailablePortFinder.getNextAvailable(1024)` then, but it is definitely good idea to log obtained port.

honza-kasik: Added `AvailablePortFinder.getNextAvailable(1024)`, port number is now logged...

rhatlapa: The port number should be guaranteed already by `java.net.ServerSocket(0)` as it tries to open server socket on that port. If it doesn't have rights to do so, such port will not be selected.

rhatlapa: => the port would by higher than 1024 if you don't have permissions for ports with lower numbers

zebrik:
LGTM. Please squash and I will merge it.

zebrik:
Thank you!

PR #113, opened 2015-11-30
Fixing failing tests in undertow subsystem

honza-kasik:
Most was caused by strange error, when command executed by dispatcher returned fail, saying that undertow server which was parent of given listener is missing. This occurred only during addition of first listener on this server. This issue looks solved when using creaper commands.

Next issue was with adding HTTPS listener. There was not any security realm configured for SSL. I finally to manage to add code for creating one and assigning it to given listener.


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/undertow/UndertowOperations.java
```diff
@@ -40,159 +46,187 @@
...
+    public void removeServletContainer(String servletContainer) throws InterruptedException, TimeoutException, IOException {
         executeRemoveAction(servletContainerTemplate.resolve(context, servletContainer));
     }
 
-    private void executeRemoveAction(ResourceAddress address) {
-        dispatcher.execute(new Operation.Builder("remove", address).build());
+    private void executeRemoveAction(ResourceAddress address) throws InterruptedException, IOException, TimeoutException {
+        Dispatcher dispatcher = new Dispatcher();
+        dispatcher.execute(new Operation.Builder("remove", address).build()); //TODO add creaper to all methods, then remove
+        dispatcher.close();
```
zebrik, line 176: I prefer to use `try` and close `Dispatcher` in `finally` block but as you mentioned this code has to be refactored to Creaper soon so I'm ok with that.

zebrik:
Thank you!

PR #114, opened 2015-12-01
Tests for PicketLink Federation subsystem

zebrik:


zebrik:
@rhatlapa please review

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/picketlink/PicketlinkOperations.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    public void disableExtension(boolean reloadImmediatellyIfRequired) throws IOException, InterruptedException, TimeoutException {
+        ops.remove(PICKETLINK_EXTENSION_ADDRESS);
+        if (reloadImmediatellyIfRequired) {
+            adminOps.reloadIfRequired();
+        }
+    }
+}
```
rhatlapa, line 56: From API point of view, I would prefer having `reloadImmediatellyIfRequired` set as part of constructor. From your usage I see that you are using `enableExtension(false)` and `disableExtension(true)`, which would mean that you would need to initialize two `PicketlinkOperations` object => which is fine from my point of view and more clear when reading the code.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/picketlink/federation/AbstractFederationTestCase.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * testcase specific substring used to distinguish between resources created by various testcases
+     */
+    protected abstract String getStringSuffix();
+
+    /**
+     * @return '{@code true}' unless system property '{@code federation.already.enabled}' is set to '{@code true}'
+     */
+    private static boolean changeSubsystem() {
+        return !Boolean.parseBoolean(ConfigUtils.get("federation.already.enabled"));
```
rhatlapa, line 113: Aren't you missing default?

rhatlapa, line 113: Looking to javadoc, it should be OK as it is.

common/src/main/java/org/jboss/hal/testsuite/fragment/WindowFragment.java
```diff
@@ -110,4 +117,13 @@ public String getTitle() {
         return title.getText();
     }
 
+    protected void clickButtonAndLogIfFails(String label) {
+        try {
+            clickButton(label);
+        } catch (WebDriverException e) {
+            log.debug("Button with label \"" + label + "\" not found");
+            throw e;
+        }
```
rhatlapa: from method name I would guess that in case of fail it will only log in case of failure. But exception is thrown as well => I suggest to use more clear name of the method.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/picketlink/PicketlinkOperations.java
```diff
@@ -0,0 +1,65 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
+
+/**
+ * Created by pjelinek on Nov 16, 2015
+ */
+public class PicketlinkOperations {
```
rhatlapa, line 37: Add javadoc.

rhatlapa:
There needs to be only done some minor improvements regarding API and javadoc, otherwise looks good.

zebrik:
@rhatlapa can you please review again?

zebrik:
I have changed last 2 commits. @rhatlapa can you please review again?

rhatlapa:
LGTM, do some minor squashing, then I will check it once more and merge.

zebrik:
Squashed a bit.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/picketlink/federation/SpAttrTestCase.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Test
+    public void strictPostBinding() throws Exception {
+        edit(InputType.CHECKBOX, spAddress, SP_STRICT_POST_BINDING, new ModelNode(false)).andVerifySuccess();
+        edit(InputType.CHECKBOX, spAddress, SP_STRICT_POST_BINDING, new ModelNode(true)).andVerifySuccess();
+    }
+
+    @Test
+    public void url() throws Exception {
+        edit(InputType.TEXT, spAddress, URL, new ModelNode("")).andVerifyFailure();
```
rhatlapa: You are trying to detect whether the URL won't be an empty String, shouldn't this check be rather against undefined value?

zebrik:
@rhatlapa please review again ;-)

zebrik:
Reload if required removed from `org.jboss.hal.testsuite.test.configuration.picketlink.federation.UtilFedInputChecker.edit` since it was a bit paranoid and moreover tested model not web console.
@rhatlapa please review for the last time.

rhatlapa:
LGTM, thanks for this huge update, merging.

PR #115, opened 2015-12-02
Quickstarts link update due to JBEAP-1591 fix.

zebrik:


zebrik:
I changed my mind and have pushed this directly.

PR #117, opened 2015-12-03
Added backup/restore command for attributes

honza-kasik:
~~Depends on creaper upstream right now.~~

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
@@ -60,9 +60,7 @@ public abstract class TransactionsTestCaseAbstract {
     public static void beforeClass() throws CommandFailedException {
         backup = new BackupAndRestoreAttributes.Builder(Address.of("subsystem", "transactions"))
                 .excluded(PROCESS_ID_SOCKET_BINDING_ATTR)
-                .dependency(USE_JDBC_STORE, "jdbc-store-datasource")
-                .dependency(USE_JDBC_STORE_ATTR, JDBC_STORE_DATASOURCE_ATTR)
-                .excluded("process-id-socket-binding")
+                .dependency(USE_JDBC_STORE, JDBC_STORE_DATASOURCE_ATTR)
                 .build();
         client.apply(backup.backup());
     }
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
@@ -67,8 +67,11 @@ public abstract class TransactionsTestCaseAbstract {
 
     @AfterClass
     public static void afterClass() throws CommandFailedException, IOException {
-        client.apply(backup.restore());
-        client.close();
+        try {
+            client.apply(backup.restore());
+        } finally {
+            client.close();
+        }
     }
 
     //helper methods
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
@@ -3,6 +3,8 @@ package org.jboss.hal.testsuite.test.configuration.transactions;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.jboss.arquillian.drone.api.annotation.Drone;
 import org.jboss.arquillian.graphene.page.Page;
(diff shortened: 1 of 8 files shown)


honza-kasik:
Ready to be merged. @zebrik Could you please review?

common/src/main/java/org/jboss/hal/testsuite/creaper/command/BackupAndRestoreAttributes.java
```diff
@@ -0,0 +1,163 @@
...
+         */
+        public Builder dependency(String attribute, String dependsOn) {
+            if (dependencies == null && dependsOn != null) {
+                dependencies = new HashMap<>();
+            }
+            if (dependencies != null) {
+                dependencies.put(attribute, dependsOn);
+            }
+            return this;
+        }
```
rhatlapa, line 131: attribute might depend only on single attribute? Isn't it possible to depend on more?

common/src/main/java/org/jboss/hal/testsuite/creaper/command/BackupAndRestoreAttributes.java
```diff
@@ -0,0 +1,163 @@
...
+            BackupAndRestoreAttributes.this.backup = ops.readResource(address, ReadResourceOption.INCLUDE_DEFAULTS, ReadResourceOption.ATTRIBUTES_ONLY).value();
+        }
+    };
+
+    private final OnlineCommand restorePart = new OnlineCommand() {
+
+        //<dependency which needs to be processed before dependent attribute is added to batch, dependent attributes>
+        private final Map<String, Set<Property>> waitingForDependency = new HashMap<>(); //waiting queue
+        private final Set<String> processedAttributes = new HashSet<>(); //attributes already in batch
+        private final Batch batch = new Batch();
```
rhatlapa, line 58: Wouldn't it be possible to use rather linked list or some priority queue to hold the dependencies? I believe it would make the code cleaner.

common/src/main/java/org/jboss/hal/testsuite/creaper/command/BackupAndRestoreAttributes.java
```diff
@@ -0,0 +1,163 @@
...
+                    } else {
+                        addToBatch(processingAttribute);
+                        addAllWaitingDependenciesToBatch(processingAttribute);
+                    }
+                }
+            }
+
+            Operations ops = new Operations(ctx.client);
+            ops.batch(batch);
+        }
```
rhatlapa, line 104: `backup` should be marked as resolved => `null`, WDYT?

honza-kasik:
I used DirectedAcyclicGraph from org.jgrapht. It adds dependency on jgrapht, but I found out, that all I was trying to do was to implement directed acyclic graph. So I used one, which was already there rather than reinventing the wheel.

zebrik:
LGTM now. Can be merged in my opinion. @rhatlapa WDYT?
May be comment  can be written that `Builder.dependency(String, String)` has higher priority than `Builder.excluded(String)` but it's probably not needed since it's quite evident that there should not be conflict between excluded and depended on attributes.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/TransactionsTestCaseAbstract.java
```diff
@@ -37,12 +44,33 @@
...
+                .excluded(PROCESS_ID_SOCKET_BINDING_ATTR)
+                .dependency(USE_JDBC_STORE, JDBC_STORE_DATASOURCE_ATTR)
+                .build();
+        client.apply(backup.backup());
+    }
+
+    @AfterClass
+    public static void afterClass() throws CommandFailedException, IOException {
+        client.apply(backup.restore());
+        client.close();
```
rhatlapa: restore might fail with exception => client should be closed in finally block

common/src/main/java/org/jboss/hal/testsuite/creaper/command/BackupAndRestoreAttributes.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * @author Jan Kasik <jkasik@redhat.com>
+ *         Created on 27.11.15.
+ */
+public class BackupAndRestoreAttributes {
```
rhatlapa: make the command classes `final` if you don't expect to be changed by inheritance => will make it easier when putting directly to creaper.

rhatlapa: Missing javadoc at class level.

rhatlapa:
finished review, generally looks good just some minor issues, see comments above.

zebrik:
In fact commands were final even before the last commit but I'm OK with the whole `BackupAndRestoreAttributes` class final as well. LGTM now.

PR #128, opened 2016-01-21
Fixing ServerManagementTestCase and cleaning up after its tests.

zebrik:
@rhatlapa can you please review?
@honza-kasik WDYT?

### DIFF
--- a/rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/RBACDomainOperations.java
+++ b/rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/RBACDomainOperations.java
@@ -1,0 +1,63 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.hal.testsuite.test.rbac;
+
+import java.io.IOException;
+
+import org.wildfly.extras.creaper.core.online.Constants;
+import org.wildfly.extras.creaper.core.online.ModelNodeResult;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.Address;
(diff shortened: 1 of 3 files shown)


rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/RBACDomainOperations.java
```diff
@@ -0,0 +1,63 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+/**
+ * A convenience for not administration operations needed in managed domain RBAC use cases.
+ */
+public final class RBACDomainOperations {
+
+    private Operations ops;
```
rhatlapa: Suggesting to have it final.

rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/ServerManagementTestCase.java
```diff
@@ -51,98 +60,106 @@ public static void afterClass() {
...
-        if (shouldSucceed) {
-            address = new ResourceAddress(new ModelNode("/host=" + host + "/server=" + server));
+        if (serverShouldBeVisible) {
+            assertEquals("Missing " + serverState.splitButtonOption + " option for server '" + server + "' of host '"
+                    + host + "'.", buttonShouldBeVisible, buttonVisibility);
 
-            verifier.verifyAttribute(address, "server-state", "running");
+            if (buttonShouldBeVisible) {
+                Address address = ops.getServerAddress(host, server);
+                new ResourceVerifier(address, client, 4000).verifyAttribute("server-state", serverState.attrValue);
```
rhatlapa: Why 4000 ms? If this is common timeout for multiple cases when using ResourceVerifier, it would make sense putting it to some constant.

rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/ServerManagementTestCase.java
```diff
@@ -51,98 +60,106 @@ public static void afterClass() {
...
         }
 
     }
 
+    private enum ServerState {
+        RUNNING("Start", "running"), STOPPED("Stop", "STOPPED");
+
+        String splitButtonOption, attrValue;
+
+        ServerState(String splitButtonOption, String attrValue) {
```
rhatlapa: Why having it names splitButtonOption? The name should clearly define what it is from user point of view (not related to the html source code as such)

rhatlapa:
Genarally LGTM, I just mentioned several suggestions for slight improvements.

honza-kasik:
Looks good, much nicer code! Thanks!

zebrik:
Suggestions worked in and force pushed.

rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/RBACDomainOperations.java
```diff
@@ -0,0 +1,63 @@
...
+
+    RBACDomainOperations(OnlineManagementClient client) {
+        ops = new Operations(client);
+    }
+
+    Address getServerAddress(String host, String server) {
+        return Address.host(host).and(Constants.SERVER, server);
+    }
+
+    boolean isServerRunning(String host, String server) throws IOException {
```
rhatlapa, line 48: shouldn't this be public? BTW good candidate for purposing for addition in creaper

rhatlapa:
Thanks, LGTM. Making it public can be incorporated any time when needed

rbac/src/test/java/org/jboss/hal/testsuite/test/rbac/RBACDomainOperations.java
zebrik, line 48: So far it's intended to be used inside this package only. I will probably propose PR to Creaper as public method and than this one will be removed.

PR #129, opened 2016-01-22
Update wildfly core dependecies.

zebrik:
@rhatlapa , @honza-kasik any complaints?


rhatlapa:
No complains, merging.

PR #126, opened 2016-01-14
Fixed messaging.connections and .clustering tests and used Creaper where possible

honza-kasik:
~~Not ready for merge yet!~~

PR #135, opened 2016-01-26
Fixed tests and added Creaper in messaging.destinations

honza-kasik:
Depends on #126

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/messaging/destinations/ConnectionFactoriesTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/messaging/destinations/ConnectionFactoriesTestCase.java
@@ -1,17 +1,13 @@
 package org.jboss.hal.testsuite.test.configuration.messaging.destinations;
 
+import org.apache.commons.lang.RandomStringUtils;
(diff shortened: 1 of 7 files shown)


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/messaging/destinations/ConnectionFactoriesTestCase.java
```diff
@@ -49,60 +45,46 @@
...
-        page.selectInTable(NAME, 0);
-        page.remove();
+    public void removeConnectionFactory() throws Exception {
+        page.remove(CONNECTION_FACTORY_TBR);
+        new ResourceVerifier(CONN_FACTORY_TBA_ADDRESS, client).verifyDoesNotExist();
+    }
 
-        verifier.verifyResource(address, false);
+    private static void createConnectionFactory(Address address, String jndiName, String connector) throws Exception {
+        operations.add(address, Values.empty()
```
zebrik, line 84: Is there any particular reason why not use `ofList` method instead of `empty` one? Not really important just may be a bit confusing.

zebrik:
The complaint above is not important. LGTM. Thx. Merging.

PR #138, opened 2016-01-27
Ability to finish form by clicking Save directly instead of waiting f…

zebrik:
…or nonexisting buttons.

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/fragment/shared/modal/WizardWindow.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/fragment/shared/modal/WizardWindow.java
@@ -28,48 +28,44 @@ public class WizardWindow extends WindowFragment {
     }
 
     /**
-     * Clicks either Finish or Save button and wait's unit the wizard window is closed
+     * Clicks Save button and waits unit the wizard window is closed.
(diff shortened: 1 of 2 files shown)


zebrik:
@honza-kasik @rhatlapa  WDYT?

common/src/main/java/org/jboss/hal/testsuite/fragment/shared/modal/WizardWindow.java
```diff
@@ -28,24 +28,57 @@ public void next() {
...
+        return isWizardClosed();
+    }
+
+    /**
+     * Clicks Save button, dismiss 'Reload required' window if appears and waits unit the wizard window is closed.
+     * @return true if window is not present after the button was clicked
+     */
+    public boolean saveAndDismissReloadRequiredWindow() {
+        clickSave();
+        dismissRealoadRequitedWindow();
```
rhatlapa, line 46: Noticed now typo in the method name `dismissRealoadRequitedWindow` => `dismissReloadRequiredWindow`. The method name should be refactored :-)

rhatlapa:
The method refactoring is not part of this MR, otherwise it is OK, merging.

PR #141, opened 2016-02-02
Infinispan fixes and obsolete APIs replacement.

zebrik:
@rhatlapa , @honza-kasik WDYT? As soon as there will be agreement on API I will continue with refactoring of remaining Infinispan tests.

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/util/ConfigChecker.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/util/ConfigChecker.java
@@ -1,0 +1,119 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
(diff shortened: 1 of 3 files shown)


zebrik:
ConfigChecker build separated to builder to enhance readability as requested.

common/src/main/java/org/jboss/hal/testsuite/util/ConfigChecker.java
```diff
@@ -0,0 +1,145 @@
...
+
+        public Builder(OnlineManagementClient client, Address resourceAddress) {
+            this.client = client;
+            this.resourceAddress = resourceAddress;
+        }
+
+        /**
+         * setter for {@link ConfigFragment}
+         */
+        public Builder config(ConfigFragment config) {
```
rhatlapa: Maybe, I would change the name to configFragment to make it more clear that the UI element is expected.

rhatlapa:
Just minor comment, otherwise LGTM

zebrik:
OK, method rename force pushed.

PR #142, opened 2016-02-15
Now creating socket binding in right socket binding group in jgroups tests

honza-kasik:
Socket binding was previously created in `full-sockets` instead of `full-ha-sockets` in domain.


zebrik:
LGTM. Thank you.

PR #147, opened 2016-02-17
Fixed failing tests in ResourceAdatersTestCase

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/page/config/ResourceAdaptersPage.java
```diff
@@ -26,24 +31,45 @@ public ResourceAdaptersConfigArea getConfigArea() {
...
+                    .addAddress(FinderNames.SUBSYSTEM, "Resource Adapters");
+
+        } else {
+            navigation = new FinderNavigation(browser, StandaloneConfigEntryPoint.class)
+                    .addAddress(FinderNames.CONFIGURATION, FinderNames.SUBSYSTEMS)
+                    .addAddress(FinderNames.SUBSYSTEM, "Resource Adapters");
+        }
+        navigation.selectColumn();
+        Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();
+    }
```
zebrik: `navigate` method content can be replaced with 

> getSubsystemNavigation("Resource Adapters").selectColumn();
> Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();

But the main objection is that this new method is never used so useless IMO.

zebrik:
LGTM now. Thank you.

PR #149, opened 2016-02-22
Stabilizating messaging tests

honza-kasik:
Trying to improve stability of messaging tests. Replaced filling classes names with existing one where possible, fixed names, replace invalid attributes with valid, better cleanup...


zebrik:
LGTM. Thank you!

PR #150, opened 2016-02-22
Stabilizating infinispan tests

honza-kasik:
Closing reload required window, removed non-catchable exceptions from tests.


rhatlapa:
Thanks, LGTM, merging.

PR #156, opened 2016-03-02
Solved cases when reload required window appeared after removing in logging tests

honza-kasik:
- replaced TimeoutException with the right one from selenium package


zebrik:
This change made `org.jboss.hal.testsuite.page.config.LoggingPage.remove` method unused so please remove it.

zebrik:
LGTM now. Thank you.

PR #161, opened 2016-03-07
Fixed transactions subsystem tests

honza-kasik:
- Shortened node identifier test value
- Fixed wrong order of tests in JournalTestCase

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
@@ -1,34 +1,34 @@
 package org.jboss.hal.testsuite.test.configuration.transactions;
 
 import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.dmr.ModelNode;
 import org.jboss.hal.testsuite.category.Shared;
-import org.jboss.hal.testsuite.dmr.Composite;
-import org.jboss.hal.testsuite.dmr.Operation;
+import org.jboss.hal.testsuite.creaper.ResourceVerifier;
+import org.jboss.hal.testsuite.util.Console;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Batch;
 
 import java.io.IOException;
+import java.util.concurrent.TimeoutException;
 
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NAME;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VALUE;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION;
-
-/**
- * @author Jan Kasik <jkasik@redhat.com>
- *         Created on 12.10.15.
- */
 @RunWith(Arquillian.class)
 @Category(Shared.class)
-//Some test might fail due to the HAL-883
 public class JournalTestCase extends TransactionsTestCaseAbstract {
 
-    //TODO prepare for JournalStore Config
     @BeforeClass
-    public static void setUp() {
-        prepareForJournalConfiguration();
+    public static void prepareForJournalConfiguration() throws IOException, TimeoutException, InterruptedException {
+        Address address = Address.subsystem("transactions");
+        Batch batch = new Batch();
+        batch.writeAttribute(address, USE_JOURNAL_STORE_ATTR, true);
+        batch.writeAttribute(address, USE_JDBC_STORE_ATTR, false);
+        operations.batch(batch);
+        administration.restartIfRequired();
+        administration.reloadIfRequired();
     }
 
     @Before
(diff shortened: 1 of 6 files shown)


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -37,35 +29,32 @@ public void before() {
...
     public void setUseJournalStoreToFalse() throws IOException, InterruptedException {
         editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, false);
     }
 
     @Test
-    public void setJournalStoreEnableAsyncIOToTrue() throws IOException, InterruptedException {
+    @InSequence(0)
+    public void toggleJournalStoreEnableAsyncIO() throws IOException, InterruptedException {
         editCheckboxAndVerify(address, JOURNAL_STORE_ENABLE_ASYNC_IO, JOURNAL_STORE_ENABLE_ASYNC_IO_ATTR, true);
-    }
```
rhatlapa, line 52: Instead of adding test ordering you should make them independent to each other.

rhatlapa:
There is just the one issue with introducing test ordering, otherwise it LGTM.

honza-kasik:
Removed test ordering, added reloadIfRequired and restartIfRequired to AfterClass in parental testcase.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -2,32 +2,25 @@
...
 @Category(Shared.class)
-//Some test might fail due to the HAL-883
 public class JournalTestCase extends TransactionsTestCaseAbstract {
 
-    //TODO prepare for JournalStore Config
     @BeforeClass
-    public static void setUp() {
+    public static void setUp() throws InterruptedException, TimeoutException, IOException {
         prepareForJournalConfiguration();
     }
```
rhatlapa, line 32: The @BeforeClass method should be directly the `prepareForJournalConfiguration` as there is nothing else done and it is good to name methods based on what they are really doing, not just encapsulate single method call (I don't see any benefit in it.)

But OK.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -37,35 +30,34 @@ public void before() {
...
+        try {
+            editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, false);
+            Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();
+            editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, true);
+        } finally {
+            dispatcher.execute(new Operation.Builder("write-attribute", address)
+                    .param("name", USE_JOURNAL_STORE)
+                    .param("value", false)
+                    .build());
+        }
```
rhatlapa, line 50: You should use creaper. Using dispatcher should be deprecated and we should be moving everything to creaper

rhatlapa, line 50: You should also return to original value => you should find it out before the test and return the value to the original one.

rhatlapa:
Also in the `TransactionsTestCaseAbstract` there are duplicate constants, one with _ATTR and one without. I believe they should be unified to single constant, don't see reason for having two of them.

zebrik:
@rhatlapa I believe the duplicated constants are there to distinguish between UI and model identifiers which are currently same but may not be in the future. But still in this case we may consider unification.

rhatlapa:
@zebrik if that is so, I believe that usage of those attributes is not fully correct (the opposite constant might be used in some cases). And also there should be the names clearly distinguished.

Also I would suggest initializing the constant based on value of the previous one till those values are the same => from API point of view those would be two constants, from implementation point of view, they would be the same.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -37,35 +30,34 @@ public void before() {
...
-    public void setUseJournalStoreToFalse() throws IOException, InterruptedException {
-        editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, false);
+    public void toggleUseJournalStore() throws IOException, InterruptedException {
+        try {
+            editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, false);
+            Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();
+            editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, true);
+        } finally {
+            dispatcher.execute(new Operation.Builder("write-attribute", address)
+                    .param("name", USE_JOURNAL_STORE)
```
rhatlapa: This is the case which seems as using incorrect constant and should be using `USE_JOURNAL_STORE_ATTR` => please use proper constants or ideally just unify them. If there will be needed distinction between them, it can be done later on.

zebrik:
@honza-kasik so please either 
- unify those constants (easier and more readable) or if you prefer
- fix their usage, initialize value of UI identifier from value of DMR identifier (or vice versa) and consider constants renaming to better distinguish between UI and DMR identifiers

honza-kasik:
@rhatlapa @zebrik Thanks for comments! Added Creaper, unified constants. @zebrik is right, the duplicity was there to distinguish UI/model identifiers. But I was too fast and now only those, which have `_ATTR` suffix are present. I hope I did not forget to fix any of your comments...

zebrik:
Oh, you were quicker ;-)

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -37,35 +36,25 @@ public void before() {
...
-        editCheckboxAndVerify(address, USE_JOURNAL_STORE, USE_JOURNAL_STORE_ATTR, true);
+    public void toggleUseJournalStore() throws Exception {
+        try {
+            page.getConfigFragment().editCheckboxAndSave(USE_JOURNAL_STORE_ATTR, true);
+            new ResourceVerifier(TRANSACTIONS_ADDRESS, client).verifyAttribute(USE_JOURNAL_STORE_ATTR, true);
+            Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();
+            page.getConfigFragment().editCheckboxAndSave(USE_JOURNAL_STORE_ATTR, false);
+            new ResourceVerifier(TRANSACTIONS_ADDRESS, client).verifyAttribute(USE_JOURNAL_STORE_ATTR, false);
+        } finally {
+            operations.writeAttribute(TRANSACTIONS_ADDRESS, USE_JOURNAL_STORE_ATTR, true);
```
zebrik: I appreciate Creaper usage here but there was yet one more objection: 

> You should also return to original value => you should find it out before the test and return the value to the original one.

zebrik:
Apart from the the way how is "use-journal-store" attr value restored all objections seem to be fixed.

honza-kasik:
Noe restoring attribute properly. Shall I squash commits?

zebrik:
Now LGTM. Yes please squash the commits.

honza-kasik:
Commits squashed!

zebrik:
Thank you!

PR #162, opened 2016-03-07
Fixed deployments tests

honza-kasik:
- Longer timeout interval
- Reload required window fixed


rhatlapa:
Thanks, LGTM, merging.

PR #164, opened 2016-03-24
Taking xml snapshots after each test class ends running

honza-kasik:
To have an overview over all created resources.

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/testlistener/TestListener.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/testlistener/TestListener.java
@@ -1,0 +1,94 @@
+package org.jboss.hal.testsuite.testlistener;
+
+import org.apache.commons.io.FileUtils;
+import org.jboss.hal.testsuite.creaper.ManagementClientProvider;
+import org.jboss.hal.testsuite.util.PropUtils;
+import org.junit.runner.Description;
+import org.junit.runner.Result;
+import org.junit.runner.notification.RunListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.wildfly.extras.creaper.core.online.ModelNodeResult;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalTime;
+
+public class TestListener extends RunListener {
+
+    private static final String XML_SNAPSHOT_DIRECTORY = PropUtils.get("snapshots.target.directory.name");
+
+    private static final Logger log = LoggerFactory.getLogger(TestListener.class);
+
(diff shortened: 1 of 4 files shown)


honza-kasik:
Put this on hold please, because I am not sure when exactly the `testStarted` method is activated. If it runs after `@BeforeClass` it should not be the right behavior we want, because the resources for currently running class are already instantiated, but we need only non-removed resources belonging to previously run test class.

common/src/main/java/org/jboss/hal/testsuite/testlistener/TestListener.java
```diff
@@ -0,0 +1,64 @@
...
+        File copyTo = File.createTempFile(PropUtils.get("snapshots.target.file.prefix"),
+                PropUtils.get("snapshots.target.file.suffix"), directory);
+        File copyFrom = getSourceFile();
+        FileUtils.copyFile(copyFrom, copyTo);
+        log.debug("Created snapshot of server configuration in '{}'.", copyTo);
+    }
+
+    private File getSourceFile() throws IOException {
+        try (OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient()) {
+            String path = new Operations(client).invoke("take-snapshot", Address.root()).value().asString();
```
rhatlapa: Use `stringValue()` instead of `.value().asString()`

common/src/main/java/org/jboss/hal/testsuite/testlistener/TestListener.java
```diff
@@ -0,0 +1,64 @@
...
+                .concat(XML_SNAPSHOT_DIRECTORY)
+                .concat(File.separator)
+                .concat(testClazz.getCanonicalName()));
+        if (!directory.exists()) {
+            if (!directory.mkdirs()) {
+                throw new IOException("Unable to create path '" + directory.toString() + "'.");
+            }
+        }
+        File copyTo = File.createTempFile(PropUtils.get("snapshots.target.file.prefix"),
+                PropUtils.get("snapshots.target.file.suffix"), directory);
```
rhatlapa, line 77: What about putting time to the name to make clear when each snapshot was created, just a suggestion.

rhatlapa, line 77: Maybe putting the time info already to directory name, to easily find out order of testcases they were run in.

zebrik:
@honza-kasik please document the new profile in README.md

common/src/main/java/org/jboss/hal/testsuite/testlistener/TestListener.java
honza-kasik, line 77: Adding time stamp to directory name, thanks for suggestion!

honza-kasik:
- Documented profile in `README.md`
- Added few fixes
- Arquillian version is now `1.0.2.Final`

rhatlapa:

honza-kasik:
Squashed commits, thanks!

rhatlapa:
Thanks, merging.

PR #166, opened 2016-04-04
Replaced adding and removing test resources in GUI by Creaper, fixed …

honza-kasik:
…reload required window

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/page/config/ELResolverPage.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/page/config/ELResolverPage.java
@@ -1,9 +1,9 @@
 package org.jboss.hal.testsuite.page.config;
 
-import org.jboss.arquillian.graphene.Graphene;
 import org.jboss.arquillian.graphene.findby.ByJQuery;
-import org.jboss.hal.testsuite.fragment.ConfigFragment;
+import org.jboss.hal.testsuite.fragment.WindowFragment;
 import org.jboss.hal.testsuite.page.home.HomePage;
+import org.jboss.hal.testsuite.util.Console;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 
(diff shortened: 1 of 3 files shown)


honza-kasik:
Ready to be merged!

common/src/main/java/org/jboss/hal/testsuite/page/config/ServerConfigurationPage.java
```diff
@@ -60,4 +45,18 @@ public ConfigFragment getWindowFragment() {
         WebElement editPanel = browser.findElement(By.className("default-window-content"));
         return  Graphene.createPageFragment(ConfigFragment.class, editPanel);
     }
+
+    public void navigate() {
+        browser.navigate().refresh();
+        Graphene.goTo(HomePage.class);
```
rhatlapa: This looks strange navigating to HomePage, when you are actually in ServerConfigurationPage. Is it really correct?

honza-kasik: I did not notice this, I will move the methods to separate class...

honza-kasik: I looked into it and it looks like there is only one usage of ServerConfigurationPage - in ELResolverTestCase. Due to the placement of Expression Resolver tool, which is placed in footer and thus available from every location, I just chose the shortest navigation possible.
IMO it would be better solution to just refactor name of this class. @rhatlapa WDYT?

zebrik: To make it more clear I suggest to either
- remove `ServerConfigurationPage` and use `HomePage` with added functionality instead or
- rename `ServerConfigurationPage` to something like `ELResolverPage` and inherit it from `HomePage` instead of `ConfigurationPage`

honza-kasik: @zebrik Thanks for suggestions! I renamed `ServerConfigurationPage` and extended `HomePage` instead of `ConfigurationPage`.

common/src/main/java/org/jboss/hal/testsuite/page/config/ELResolverPage.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class ELResolverPage extends HomePage {
+
+    public void openExpressionResolver() {
+        WebElement tools = browser.findElement(ByJQuery.selector("div.gwt-HTML.footer-link:contains(Tools)"));
+        tools.click();
+        browser.findElement(By.xpath("//div[@class='popupContent']//a[contains(text(), 'Expression Resolver')]")).click();
+    }
+
+    public ConfigFragment getWindowFragment() {
```
zebrik: It seems that this method is now called just form `resolveSystemProperty(String name)` method of the same object thus I'm suggesting to either 
- make `getWindowFragment()` private or 
- get rid of `getWindowFragment()` method and move it's code inside of `resolveSystemProperty(String name)` method.

zebrik:
Apart from one minor API remark LGTM.

honza-kasik:
Thanks, removed `getWindowFragment()`

zebrik:
LGTM. Thank you.

PR #167, opened 2016-04-06
Extracted AvailablePortFinder from Apache Mina to reduce dependecies …

honza-kasik:
…and simplify port obtaining

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
@@ -1,0 +1,173 @@
+package org.jboss.hal.testsuite.util;
+
+import java.io.IOException;
+import java.net.DatagramSocket;
+import java.net.ServerSocket;
+import java.util.NoSuchElementException;
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * Finds currently available server ports.
+ *
+ * @author <a href="http://mina.apache.org">Apache MINA Project</a>
+ * @see <a href="http://www.iana.org/assignments/port-numbers">IANA.org</a>
+ */
+public class AvailablePortFinder {
+    /**
+     * The minimum number of user port number.
+     */
+    public static final int MIN_PORT_NUMBER = 1024;
+
+    /**
+     * The maximum number of user port number.
+     */
+    public static final int MAX_PORT_NUMBER = 49151;
+
+    /**
+     * Creates a new instance.
+     */
+    private AvailablePortFinder() {
+        // Do nothing
+    }
+
+    /**
+     * Returns the {@link Set} of currently available port numbers
+     * ({@link Integer}).  This method is identical to
+     * <code>getAvailablePorts(MIN_PORT_NUMBER, MAX_PORT_NUMBER)</code>.
+     *
+     * WARNING: this can take a very long time.
+     */
+    private static Set<Integer> getAvailablePorts() {
+        return getAvailablePorts(MIN_PORT_NUMBER, MAX_PORT_NUMBER);
+    }
+
+    /**
+     * Gets an available port, selected by the system.
+     *
+     * @throws NoSuchElementException if there are no ports available
+     */
+    private static int getNextAvailable() {
+        ServerSocket serverSocket = null;
+
+        try {
+            // Here, we simply return an available port found by the system
+            serverSocket = new ServerSocket(0);
+            int port = serverSocket.getLocalPort();
+
+            // Don't forget to close the socket...
+            serverSocket.close();
+
+            return port;
+        } catch (IOException ioe) {
+            throw new NoSuchElementException(ioe.getMessage());
+        }
+    }
+
+    /**
+     * Gets the next available port starting at a port.
+     *
+     * @param fromPort the port to scan for availability
+     * @throws NoSuchElementException if there are no ports available
+     */
+    private static int getNextAvailable(int fromPort) {
+        if (fromPort < MIN_PORT_NUMBER || fromPort > MAX_PORT_NUMBER) {
+            throw new IllegalArgumentException("Invalid start port: " + fromPort);
+        }
+
+        for (int i = fromPort; i <= MAX_PORT_NUMBER; i++) {
+            if (available(i)) {
+                return i;
+            }
+        }
+
+        throw new NoSuchElementException("Could not find an available port " + "above " + fromPort);
+    }
+
+    /**
+     * Checks to see if a specific port is available.
+     *
+     * @param port the port to check for availability
+     */
+    private static boolean available(int port) {
+        if (port < MIN_PORT_NUMBER || port > MAX_PORT_NUMBER) {
+            throw new IllegalArgumentException("Invalid start port: " + port);
+        }
+
+        ServerSocket ss = null;
+        DatagramSocket ds = null;
+
+        try {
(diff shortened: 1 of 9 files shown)


common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,168 @@
...
+    public static int getNextAvailable() {
+        ServerSocket serverSocket = null;
+
+        try {
+            // Here, we simply return an available port found by the system
+            serverSocket = new ServerSocket(0);
+            int port = serverSocket.getLocalPort();
+
+            // Don't forget to close the socket...
+            serverSocket.close();
```
rhatlapa: what about closing in case of `serverSocket.getLocalPort` failing?

rhatlapa:
As finding free port for UDP means checking port by port, on the other hand for TCP it can be retrieved very easily by just initializing ServerSocket(0), I suggest making it two different methods, one for UDP and one for TCP, you can then have common method using those two.

honza-kasik:
I modified `getNextAvailableUserPort()` method to be potentially faster by letting system first give me an available TCP port number and then check if both UDP and TCP ports are available. If the port is not available, it will still walk through ports to find an available one.

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
honza-kasik: If it fails it just returns -1. So it closes anyway...

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Returns the {@link Set} of currently avaliable port numbers ({@link Integer})
+     * between the specified port range.
+     *
+     * @throws IllegalArgumentException if port range is not between
+     * {@link #MIN_PORT_NUMBER} and {@link #MAX_PORT_NUMBER} or
+     * <code>fromPort</code> if greater than <code>toPort</code>.
+     */
+    private static Set<Integer> getAvailablePorts(int fromPort, int toPort) {
```
rhatlapa: This should be renamed as it is looking only for TCP ports.

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,173 @@
...
+            } finally {
+                if (s != null) {
+                    try {
+                        s.close();
+                    } catch (IOException e) {
+                        /* should not be thrown */
+                    }
+                }
+            }
+        }
```
rhatlapa, line 147: What about using try with resources as the code requires JDK 1.8 =>

```
        try (ServerSocket ss = new ServerSocket(i)) {
            ...
        }
```

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            if (ss != null) {
+                try {
+                    ss.close();
+                } catch (IOException e) {
+                    /* should not be thrown */
+                }
+            }
+        }
```
rhatlapa, line 115: What about using try with resources to make the code shorter and more readable?

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * The maximum number of user port number.
+     */
+    public static final int MAX_PORT_NUMBER = 49151;
+
+    /**
+     * Creates a new instance.
+     */
+    private AvailablePortFinder() {
```
rhatlapa, line 30: The javadoc is incorrect. This constructor actually prevents creating instance of this class.

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,173 @@
...
+    private static Set<Integer> getAvailablePorts() {
+        return getAvailablePorts(MIN_PORT_NUMBER, MAX_PORT_NUMBER);
+    }
+
+    /**
+     * Gets an available port, selected by the system.
+     *
+     * @throws NoSuchElementException if there are no ports available
+     */
+    private static int getNextAvailable() {
```
rhatlapa: Should be named to clearly state that it looks only for TCP ports

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,173 @@
...
+
+        throw new NoSuchElementException("Could not find an available port " + "above " + fromPort);
+    }
+
+    /**
+     * Checks to see if a specific port is available.
+     *
+     * @param port the port to check for availability
+     */
+    private static boolean available(int port) {
```
rhatlapa: availability is not very good name. You can be looking for it from different point of views and from one client PoV available can be understood as port is in use and is available to connect to it remotely. I suggest using more descriptive name to prevent confusion.

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
rhatlapa: Ok, thanks for explanation

rhatlapa:
I have mainly some API objections connected to naming to make clear what each method does.

honza-kasik:
Fixed API issues, updated to Java 8, replaced `available` for `unused`/`free to use`.

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,136 @@
...
+                // Do nothing
+            }
+        }
+        return result;
+    }
+
+    /**
+     * Gets an not used user port.
+     */
+    public static int getNextAvailableUserPort() {
```
rhatlapa, line 154: I would suggest to put to javadoc, that it checks both TCP and UDP ports

common/src/main/java/org/jboss/hal/testsuite/util/AvailablePortFinder.java
```diff
@@ -0,0 +1,136 @@
...
+
+        throw new NoSuchElementException("Could not find an unused port " + "above " + fromPort);
+    }
+
+    /**
+     * Checks to see if a specific port is not currently used.
+     *
+     * @param port the port to check for usage
+     */
+    private static boolean isPortFreeToUse(int port) {
```
rhatlapa, line 84: Also suggesting to add info to javadoc that it checks both TCP and UDP ports

rhatlapa, line 84: Also this should be probably split to two methods, one for TCP and one UDP, see usages, where you are using this method in context when you are actually want to check only TCP ones.

honza-kasik:
Split method for checking port usability to two separate methods, fixed javadoc.

rhatlapa:
Thanks, LGTM, merging.

PR #168, opened 2016-04-07
Fixed port range error in AvailablePortFinder

honza-kasik:
Due to the fact, that the `new ServerSocket(0)` allocates port in ephemeral range, I had to set higher `MAX_PORT_NUMBER` and refactor `getNextAvailableUserPort()` method.


rhatlapa:
Thanks, LGTM, merging.

PR #170, opened 2016-04-13
Removed test dependency, fixed reload required window and added screa…

honza-kasik:
…per in SocketBindingsTestCase


zebrik:
LGTM. Thank you.

PR #174, opened 2016-04-14
Added Creaper to JCASubsystemTestCase, simplified

honza-kasik:
Tried to cleanup better, but apparently omitting of boolean attribute in xml configuration is equal to `attribute="false"`.


zebrik:
LGTM. Thx.

PR #176, opened 2016-04-20
RunListener to check server state cleanup.

zebrik:


zebrik:
@rhatlapa can you please look at this?

zebrik:
@honza-kasik WDYT?

honza-kasik:
@zebrik I like it! Looks good to me! :+1:

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Created by pjelinek on Apr 19, 2016
+ */
+public class ServerStateListenerOperations {
+
+    /**
+     * Has to be initialized in every public method which requires it with new {@link OnlineManagementClient}
+     * which is closed before the method finishes.
+     */
+    private Operations ops;
```
rhatlapa: If I am looking correctly, you are using the `ops` to actually something which is provided between methods, correct?

zebrik:
Added javadoc and changed implementation of ServerStateListenerOperations.

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    private class ModelReader {
+
+        private ModelReader(OnlineManagementClient client) {
+            this.ops = new Operations(client);
+        }
+
+        private Operations ops;
```
rhatlapa: should be final. And probably before constructor.

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
```diff
@@ -0,0 +1,124 @@
...
+
+        private Stream<String> getHostStream() {
+            ModelNodeResult hostsResult = null;
+            try {
+                hostsResult = ops.readChildrenNames(Address.root(), HOST);
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to read hosts.", e);
+            }
+            hostsResult.assertDefinedValue();
+            return hostsResult.stringListValue().stream();
```
rhatlapa, line 98: should be inside the try catch block. Even though in this case catch block throws exception, so it shouldn't cause any harm.

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
```diff
@@ -0,0 +1,124 @@
...
+        private Stream<String> getServerStream(String host) {
+            ModelNodeResult serversResult;
+            try {
+                serversResult = ops.readChildrenNames(Address.host(host), SERVER);
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to read servers of host.", e);
+            }
+            serversResult.assertDefinedValue();
+            return serversResult.stringListValue().stream();
+        }
```
rhatlapa, line 110: The same as above

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
```diff
@@ -0,0 +1,124 @@
...
+
+        private String readServerConfigStatus(String host, String server) {
+            ModelNodeResult serverConfigStatusResult;
+            try {
+                serverConfigStatusResult = ops.readAttribute(Address.host(host).and(SERVER_CONFIG, server), STATUS);
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to read server config status.", e);
+            }
+            serverConfigStatusResult.assertDefinedValue();
+            return serverConfigStatusResult.stringValue();
```
rhatlapa, line 120: The same as above

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
```diff
@@ -0,0 +1,124 @@
...
+            serversResult.assertDefinedValue();
+            return serversResult.stringListValue().stream();
+        }
+
+        private String readServerConfigStatus(String host, String server) {
+            ModelNodeResult serverConfigStatusResult;
+            try {
+                serverConfigStatusResult = ops.readAttribute(Address.host(host).and(SERVER_CONFIG, server), STATUS);
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to read server config status.", e);
```
rhatlapa, line 117: Any reason for changing this to RuntimeException? Is it because of using in lambda?

zebrik, line 117: Yes it's done this way to avoid the necessity of polluting lambda inside `Function` instance with try catch block which would be necessary otherwise since `apply` method of `Function` don't allow to rethrow any checked exception.

common/src/main/java/org/jboss/hal/testsuite/testlistener/ServerStateListenerOperations.java
zebrik, line 98: The try-catch block is here just to change type of potential exception to runtime to avoid unnecessary try-catch block inside lambda (making the lambda with most of the business logic hard to read). I don't want to avoid exceptions just want to avoid checked ones to keep it easy to read. This way it's causing no harm and it is easy to read IMO.

zebrik:
Added `final` modifier to `org.jboss.hal.testsuite.testlistener.ServerStateListenerOperations.ModelReader.ops`.

zebrik:
Please don't merge this yet.

zebrik:
Added reading of init server state map only for domain mode.

rhatlapa:
Thanks, merging.

PR #177, opened 2016-04-21
Add ServerStateListener to be optionally run.

zebrik:
and rename `snapshot` profile as well as `TestListener` class

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,10 @@ Can be one of those:
 * `-Pbasic,domain` ... run basic tests against domain mode
 * `-Prbac,standalone` ... run RBAC related tests against standalone mode
 * `-Prbac,domain` ... run RBAC related tests against domain mode
-* `-P[basic,rbac],[domain,standalone],snapshots` ... run tests against given mode and take snapshot of server xml configuration file at the end of every test class run
+* `-P[basic,rbac],[domain,standalone],serverCheck` ... run tests against given mode and at the end of every test class run
+    * take snapshot of server xml configuration file
+    * check if server requires reload or restart
+    * in domain mode check if running states of servers don't change after test class run
 
 ### Required jboss.dist parameter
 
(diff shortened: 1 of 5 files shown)


zebrik:
@rhatlapa , @honza-kasik WDYT?

honza-kasik:
@zebrik Definitely good idea! LGTM!

common/src/main/java/org/jboss/hal/testsuite/testlistener/SnapshotListener.java
```diff
@@ -17,11 +17,11 @@
 import java.io.IOException;
 import java.time.LocalTime;
 
-public class TestListener extends RunListener {
+public class SnapshotListener extends RunListener {
```
rhatlapa: I would use `XmlSnapshotListener` or `ServerConfigSnapshotListener`. As Snapshot generally could be easily missinterpreted as something related to UI screenshots.

zebrik: What about `ConfigSnapshotTaker`? And rename `ServerStateListener` -> `ServerStateReporter` as well? WDYT?

rhatlapa: +1 for the `ConfigSnapshotTaker`. -1 for `ServerStateReporter` as reporter is something with special meaning, e.g. to have html report. I would vote for second name to be something like => `ServerStateAfterTestLogger`

zebrik: I was already thinking about `ServerStateLogger` as an alternative to `ServerStateReporter`. I personally don't like the `AfterTest` part since it's exposing details such if it will be run after testcase or atomic test which may change in the future.

zebrik: Renamed:
`SnapshotListener` -> `ConfigSnapshotTaker`,
`ServerStateListener` -> `ServerStateLogger`,
`ServerStateListenerOperations` -> `ServerStateLoggerOperations`

rhatlapa:
Thanks, LGTM, merging

PR #179, opened 2016-05-24
Undertow metrics

rhatlapa:
Updated and improved coverage of undertow metrics to properly check that the metrics values are also updated in console.


pom.xml
```diff
@@ -98,6 +98,13 @@
         <version>${version.org.jgrapht}</version>
       </dependency>
 
+      <!-- servlet dependencies -->
+      <dependency>
+        <groupId>org.jboss.spec.javax.servlet</groupId>
+        <artifactId>jboss-servlet-api_3.0_spec</artifactId>
```
zebrik: Just curious why `jboss-servlet-api_3.0_spec` is used and not `jboss-servlet-api_3.1_spec` ?
Nevertheless this is probably of no importance in this case.

rhatlapa: I used jboss-servlet-api_3.0_spec as it is also available for EE6, but you are right, I could have used 3.1 which is used with EE7 and the tests are written for hal used with WildFly 10+ and EAP 7 which both implement EE7. As I am not using any Servlet 3.1 specific functionalities, I would say, it is OK to leave it as it is.

rhatlapa: Looking more, I should probably change the scope to `provided` as in the end the specification available in the EAP/WildFly shall be used.

rhatlapa: Updated to 3.1 and changed scope to provided.

rhatlapa:
I have updated and squashed the commits at the same time as the only change I've done is the one in pom.xml

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
```diff
@@ -1,131 +1,253 @@
...
+ * Real values are generated to see whether they are really updated in the web console.
  */
 @RunWith(Arquillian.class)
+@RunAsClient
 @Category(Shared.class)
 public class WebMetricsTestCase {
 
+    private static final Logger log = LoggerFactory.getLogger(WebMetricsTestCase.class);
+
+    private static final String DEPLOYMENT_NAME = "servlets";
```
zebrik: May be `DEPLOYMENT_NAME` could be more specific just in case that something breaks cleaning up in the future and we want to figure out where the deployment comes from. Not much important though.

zebrik:
I wrote one last optional suggestion related to deployment name. You can take it into account or not. It's ready to be merged.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/WebMetricsTestCase.java
rhatlapa: Do you have any suggestion for better name?

zebrik: `metrics_generating_servlets` or may be even better `WebMetricsTestCase_servlets` ?

rhatlapa: Thanks for the suggestions, I've used the second variant. Already commit updated and squashed.

zebrik:
Thank you.

PR #180, opened 2016-05-30
Optimizing serverCheck profile logging output.

zebrik:
Adding testcase context to the ServerStateLogger log output.
Removing unnecessary logs from ConfigSnapshotTaker log output.
Making test listeners code easier to read.


rhatlapa:
LGTM, merging.

PR #190, opened 2016-09-13
Changing profile for Infinispan subsystem configuration testing to full-ha

zebrik:
Many of Infinispan subsystems features doesn't work without JGroups and it's not any more possible to add those resources without JGroups subsystem being configured.


rhatlapa:
Thanks, LGTM, merging.

PR #193, opened 2016-09-14
HAL-1149 WS deployment enpoint names change.

zebrik:


rhatlapa:
Interesting, that the colon doesn't need to be any longer encoded :-). Merging.

zebrik:
It was intentional fix https://issues.jboss.org/browse/HAL-1149
https://github.com/hal/core/pull/279/

PR #197, opened 2016-09-19
Added reload when needed in ConfigurationChangesTestCase + several tweaks

honza-kasik:
- `assertSuccess()` on write max-history attribute operation
- max-history value as constant
- added reload after write-atribute/add in `before()`

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/runtime/server/configurationchanges/ConfigurationChangesTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/runtime/server/configurationchanges/ConfigurationChangesTestCase.java
@@ -52,15 +52,14 @@ public class ConfigurationChangesTestCase {
     @Page
     private ConfigurationChangesPage page;
 
-    private static OnlineManagementClient client;
-    private static Operations operations;
+    private static final OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient();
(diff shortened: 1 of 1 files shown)


honza-kasik:
I will squash it, wait

honza-kasik:
Done!

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/server/configurationchanges/ConfigurationChangesTestCase.java
```diff
@@ -52,15 +52,14 @@
...
+    private static final Operations operations = new Operations(client);
+    private static final Administration administration = new Administration(client);
     private static Address serviceAddress;
     private static ConfigurationChangesResource changesResource;
 
     @BeforeClass
-    public static void beforeClass() {
-        client = ManagementClientProvider.createOnlineManagementClient();
-        operations = new Operations(client);
+    public static void beforeClass() throws InterruptedException, TimeoutException, IOException {
```
rhatlapa: No need for the throws as those exceptions are not thrown from the used methods.

rhatlapa (changes requested):
No need for the throws ... in the `beforeClass`

honza-kasik:
Removed exceptions from before class, thanks!

rhatlapa:
Thanks, LGTM now, merging.

PR #202, opened 2016-10-05
HAL-1188: JDBC tab changed to Store tab in transactions subsystem

zebrik:


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -1,62 +1,67 @@
...
+        }
+    }
+
+    private void navigate2store() {
+        page.navigate();
+        page.getConfig().switchTo("Store");
+    }
+
+    private boolean isUseJournalStore() throws IOException {
+        return operations.readAttribute(TRANSACTIONS_ADDRESS, USE_JOURNAL_STORE_ATTR).value().asBoolean();
```
rhatlapa: use `.booleanValue()` instead of `.value().asBoolean()`

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -1,62 +1,67 @@
...
+        boolean originalUseJournalStoreValue = isUseJournalStore();
         try {
             page.getConfigFragment().editCheckboxAndSave(USE_JOURNAL_STORE_ATTR, true);
             new ResourceVerifier(TRANSACTIONS_ADDRESS, client).verifyAttribute(USE_JOURNAL_STORE_ATTR, true);
             Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();
             page.getConfigFragment().editCheckboxAndSave(USE_JOURNAL_STORE_ATTR, false);
             new ResourceVerifier(TRANSACTIONS_ADDRESS, client).verifyAttribute(USE_JOURNAL_STORE_ATTR, false);
         } finally {
-            operations.writeAttribute(TRANSACTIONS_ADDRESS, USE_JOURNAL_STORE_ATTR, value);
+            writeUseJournalStore(originalUseJournalStoreValue);
```
rhatlapa, line 28: The proper way to restore the value would be by reading the result as `ModelNode`, which can be also `undefined` and then restore the `ModelNode`.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
```diff
@@ -1,62 +1,67 @@
...
+        page.navigate();
+        page.getConfig().switchTo("Store");
+    }
+
+    private boolean isUseJournalStore() throws IOException {
+        return operations.readAttribute(TRANSACTIONS_ADDRESS, USE_JOURNAL_STORE_ATTR).value().asBoolean();
+    }
+
+    private void writeUseJournalStore(boolean value) throws IOException, InterruptedException, TimeoutException {
+        if (isUseJournalStore() != value) {
```
rhatlapa, line 60: Handling for performance reasons? :-) Is it really needed, does the transaction subsystem require reload even when the value wasn't changed?

rhatlapa (commented):
Just minor issues. when incorporated, you can squash the commits and merge this PR.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
zebrik: Actually it requires even restart. Even when the value is not changed. Really.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
zebrik: +1 Thank you.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
rhatlapa: Interesting, I know that for some subsystems the model is able to detect whether the value was actually changed and do basically this check internally. => Thanks for the info and definitely worth to have this check here.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/transactions/JournalTestCase.java
zebrik: I understand your point and agree in general but this attribute can not be undefined and it would force me to make the code less readable so I will not do it.

rhatlapa: Sure, no issue on my side. Just when I notice something, I add comment. Definitely this doesn't block this PR from merging.

zebrik:
Thank you for review. I have force pushed update and am going to merge this.

PR #206, opened 2016-10-17
Added Travis CI status icon to README

honza-kasik:


rhatlapa:
I didn't want to add the travis icon here as the travis doesn't run any tests now => it could be misinterpreted as everything is passing even though in our case it just checks that checkstyle passed.

@zebrik @honza-kasik if you want to add it even with the risk of this misinterpretation, I shall merge it.

zebrik:
IMO it might make sense with some smoke tests run on Trevis. Without any tests it's just cool but without any real benefit. 
But something like smoke test profile should be useful in other scenarios as well (multiplatform testing etc.)

rhatlapa (approved):

PR #208, opened 2016-10-31
Removed failing message for HAL-1169 since it was resolved

honza-kasik:


rhatlapa:
The link to the original issue might still be useful, e.g. when running the test against some version of WildFly/EAP with older version of HAL not containing the fix => rejecting this change, acceptable would be if there would be message saying, should be fixed since ... version, with older versions it fails probably due...

PR #209, opened 2016-10-31
Fixed test which failed due to chages in model

honza-kasik:


rhatlapa:
Thanks, LGTM, merging.

PR #214, opened 2016-11-14
[HAL-652] Tests for datasource templates

honza-kasik:
Just smoke tests for now, need to resolve driver detection on domain.

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/DatasourceTemplatesTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/DatasourceTemplatesTestCase.java
@@ -1,0 +1,159 @@
+package org.jboss.hal.testsuite.test.configuration.datasources;
+
+import org.jboss.arquillian.drone.api.annotation.Drone;
+import org.jboss.arquillian.graphene.page.Page;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.hal.testsuite.category.Shared;
+import org.jboss.hal.testsuite.fragment.config.datasource.DatasourceWizard;
+import org.jboss.hal.testsuite.fragment.formeditor.Editor;
+import org.jboss.hal.testsuite.page.config.DatasourcesPage;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.openqa.selenium.WebDriver;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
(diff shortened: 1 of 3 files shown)


honza-kasik:
I filled https://issues.jboss.org/browse/HAL-1228 and I will add more tests when it will be clear how it will be resolved. Consider this PR final and ready to be merged!

zebrik:
I'm missing MariaDB - should be added as a template.

honza-kasik:
MariaDB template added in https://issues.jboss.org/browse/HAL-1230

zebrik:
I see, it's already merged. Congrats ;-) So IMO we can already add it to `DatasourceTemplatesTestCase.DatasourceTemplate`.

honza-kasik:
Thanks! I added the template to the test.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/DatasourceTemplatesTestCase.java
```diff
@@ -0,0 +1,159 @@
...
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Contains smoke tests which verifies that wizard in datasources subsystem contains correctly predefined values for
+ * each template.
+ */
+@Category(Domain.class)
```
zebrik: Why is tested only domain mode? In scenarios like this where could be differences between behaviour I would suggest to test both modes.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/DatasourceTemplatesTestCase.java
```diff
@@ -0,0 +1,159 @@
...
+        private String driverClass;
+
+        DatasourceTemplate(String datasourceTemplateName, String name, String moduleName, String driverClass) {
+            this.datasourceTemplateName = datasourceTemplateName;
+            this.name = name;
+            this.moduleName = moduleName;
+            this.driverClass = driverClass;
+        }
+
+        public String getDatasourceTemplateName() {
```
zebrik, line 69: Why this and following methods are public? Just curious - not so important.

honza-kasik: I know that it would make sense to mark them private, but how do I then 
distinguish methods which should be used in parent class and methods that does not make sense to use in parent class?

honza-kasik:
Changed `Domain.class` to `Shared.class` and now it runs in standalone too

zebrik:
Please squash and I will merge.

zebrik:
Thank you. Merging...

PR #215, opened 2016-11-22
Fixed RequestControllerSubsystemTestCase

honza-kasik:
* Fixed bugs in `toggleTrackIndividualEndpoints`
* Fixed attributes presence based on server mode, see https://github.com/wildfly/wildfly-core/blob/5ff6433a5f06051ce8c29c654e9957d463e9ce77/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestControllerExtension.java and `private static Collection<AttributeDefinition> getAttributeDefinitions(boolean registerRuntimeOnly)` in https://github.com/wildfly/wildfly-core/blob/5ff6433a5f06051ce8c29c654e9957d463e9ce77/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestControllerRootDefinition.java

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/requestcontroller/RequestControllerSubsystemTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/requestcontroller/RequestControllerSubsystemTestCase.java
@@ -89,22 +89,25 @@ public class RequestControllerSubsystemTestCase {
 
     @Test
     public void toggleTrackIndividualEndpoints() throws Exception {
-        ModelNodeResult originalValue = operations.readAttribute(REQUEST_CONTROLLER_SUBSYSTEM_ADDRESS,
+        final ModelNodeResult originalValue = operations.readAttribute(REQUEST_CONTROLLER_SUBSYSTEM_ADDRESS,
                 TRACK_INDIVIDUAL_ENDPOINTS, ReadAttributeOption.NOT_INCLUDE_DEFAULTS);
+        final ModelNodeResult defaultValue = operations.readAttribute(REQUEST_CONTROLLER_SUBSYSTEM_ADDRESS,
+                TRACK_INDIVIDUAL_ENDPOINTS, ReadAttributeOption.INCLUDE_DEFAULTS);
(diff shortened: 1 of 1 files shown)


honza-kasik:
I reviewed the possibility to have `@Category` for class, but it doesn't work with `@Category(Shared.class)` for class and `@Category(Standalone.class)` for one test - it runs all of them in domain mode and if I would revert it it would be IMHO less clear which test runs when...

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/requestcontroller/RequestControllerSubsystemTestCase.java
```diff
@@ -59,6 +59,7 @@ public static void afterClass() throws InterruptedException, TimeoutException, I
         }
     }
 
+    @Category(Standalone.class)
     @Test
     public void verifyActiveRequests() throws Exception {
```
rhatlapa, line 64: This test can be updated also for domain, the runtime attributes are specific for single server => you need to check it on the server level and not on profile level => all tests can remain shared for standalone and domain, just domain requires handling for single server in domain rather then whole profile.

rhatlapa (changes requested):

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/requestcontroller/RequestControllerSubsystemTestCase.java
honza-kasik: Maybe I am not looking everywhere but I do not see a way how to see active request on server in web console gui. That looks like a possible bug.

rhatlapa: Yes, I also believe it is a bug, I don't see it there either.

rhatlapa (approved):

rhatlapa:
LGTM, merging. Note, there should be added test for the `active-requests` in domain even though the functionality is currently missing in HAL

PR #219, opened 2016-12-01
Added tests for prepared transactions

honza-kasik:
Thee location of prepared transactions view will be changed in one of upcoming builds, so the navigation has to be updated afterwards.


transactions/src/test/java/org/jboss/hal/testsuite/test/transactions/TransactionCommitAndRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+
+    private static String[] TRANSATIONS_XID = new String[]{
+            "AAAAAAAAAAAAAP__CigFZuOgfONXxVvAAAASdQAAAAEAAAAAAAAAAAAAAAAAAP__CigFZuOgfONXxVvAAAASdDEHAgIA",
+            "AAAAAAAAAAAAAP__CigFZuOgfONXxVvAAAASvwAAAAEAAAAAAAAAAAAAAAAAAP__CigFZuOgfONXxVvAAAASdDEHAgIA",
+            "AAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFpgAAAAEAAAAAAAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFSTEHAgIA",
+            "AAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFSgAAAAEAAAAAAAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFSTEHAgIA",
+            "AAAAAAAAAAAAAP__CigFZuOgfONXxVvAAAASbwAAAAEAAAAAAAAAAAAAAAAAAP__CigFZuOgfONXxVvAAAASbjEHAgIA",
+            "AAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFrgAAAAEAAAAAAAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFrTEHAgIA",
+            "AAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFsgAAAAEAAAAAAAAAAAAAAAAAAP__CigFZtXWO0pXxVwMAAAFrTEHAgIA"
+    };
```
rhatlapa, line 100: Please add comment explaining what the values mean, how they were counted. I guess they are based on the provided transaction journal. If it is so, add note about it.

transactions/src/test/java/org/jboss/hal/testsuite/test/transactions/TransactionCommitAndRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+        if (errors.size() > 0) {
+            throw new AssertionError(errors.stream().map(AssertionError::getMessage).collect(Collectors.joining(", ")));
+        }
+    }
+
+    private static final class LogListenerForRollbackAndCommit extends TailerListenerAdapter {
+
+        public LogListenerForRollbackAndCommit() {
+        }
```
rhatlapa: Empty constructor exists by default if there is no other constructor if I am not mistaken, if it is so, please remove it.

transactions/src/test/resources/logback.xml
```diff
@@ -0,0 +1,15 @@
...
+            <!-- pattern format: http://logback.qos.ch/manual/layouts.html#conversionWord -->
+        </encoder>
+    </appender>
+
+    <logger name="org.jboss.hal.testsuite" level="DEBUG"/>
+
+    <root level="debug">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
```
rhatlapa: logback-test is enough, you don't need also loback.xml

rhatlapa (commented):
Rather minor issues which mainly improves documenting of what is being done. When the notes are added, it can be merged.

honza-kasik:
Worked in suggestions, improved README, thanks!

transactions/src/test/java/org/jboss/hal/testsuite/test/transactions/TransactionCommitAndRollbackTestCase.java
```diff
@@ -85,7 +85,10 @@ public void before() {
         page.navigate();
     }
 
-    private static String[] TRANSATIONS_XID = new String[]{
+    /**
+     * Unique identifiers of prepared transactions as they appear in UI
+     */
+    private static String[] TRANSACTIONS_XID = new String[]{
```
rhatlapa, line 92: Still missing important information which is that the IDs correspond to the journal in the resources => when different journal would be used, these would have to be updated. If I don't understand it correctly, please correct me.

rhatlapa (changes requested):

transactions/src/test/java/org/jboss/hal/testsuite/test/transactions/TransactionCommitAndRollbackTestCase.java
honza-kasik: Well, since the journal is just a bunch of binary files, I did not extract ids from it, but I used ids displayed in Web Console instead. For future updates, it needs to be taken from there again after the prepared journal is installed.

honza-kasik:
Updated docs and squashed. Thanks!

rhatlapa (approved):

rhatlapa:
LGTM, thanks, merging.

PR #220, opened 2016-12-05
Removing edit of deprecated attributes in HTTPS Listener.

zebrik:
HTTP listener attributes ssl-session-cache-size and ssl-session-timeout are not any more configurable
in Undertow configuration in webconsole. Instead they should be configured on the Elytron security context.


honza-kasik:
I filled bug https://issues.jboss.org/browse/HAL-1248

PR #224, opened 2016-12-16
Updated parent dependency in transactions module

honza-kasik:


zebrik:
Thank you, merging...

PR #227, opened 2016-12-19
Fixed navigation in DeploymentsPage

honza-kasik:


zebrik:
Thank you. Merging ...

PR #228, opened 2016-12-19
Fixed navigation in transaction module

honza-kasik:
Fixed navigation for next hal update (2.9.0.Final) and now handling AssertionErrors better in test.


common/src/main/java/org/jboss/hal/testsuite/util/AssertionErrorCollector.java
```diff
@@ -0,0 +1,62 @@
...
+        final StringBuilder builder = new StringBuilder();
+
+        builder.append("Test failed in ").append(amountOfErrors()).append(" case").append(amountOfErrors() > 1 ? "s" : "").append(". \n");
+        for (Throwable error : assertionErrors) {
+            builder.append(error.getMessage()).append('\n');
+        }
+        return builder.toString();
+    }
+
+}
```
zebrik: Please add new line at the end of file.

common/src/main/java/org/jboss/hal/testsuite/util/AssertionErrorCollector.java
```diff
@@ -0,0 +1,62 @@
...
+    public int amountOfErrors() {
+        return assertionErrors.size();
+    }
+
+    private String createErrorMessage() {
+        final StringBuilder builder = new StringBuilder();
+
+        builder.append("Test failed in ").append(amountOfErrors()).append(" case").append(amountOfErrors() > 1 ? "s" : "").append(". \n");
+        for (Throwable error : assertionErrors) {
+            builder.append(error.getMessage()).append('\n');
```
zebrik: I'm missing stack trace.

common/src/main/java/org/jboss/hal/testsuite/util/AssertionErrorCollector.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Collects all assertion errors which occurred during test execution
+ * to report them back at the end of test lifecycle. This approach
+ * allows to run through entire test execution and executed all
+ * required phases (like cleaning up database at the end of each
+ * persistence test) - so called soft assertion.
+ *
+ * @author <a href="mailto:bartosz.majsak@gmail.com">Bartosz Majsak</a>
+ */
+public class AssertionErrorCollector {
```
zebrik: @rhatlapa please look at this class as well.

common/src/main/java/org/jboss/hal/testsuite/util/AssertionErrorCollector.java
honza-kasik: Is having a stack trace here necessary? My point is that I know which tests failed and the fail message should be sufficient. On the other hand, it could be useful in case of debugging, although there would have to be other assertion in method call queue to catch irrelevant assertion...

zebrik: It depends on expected usage of this class, we should discuss it f2f.

common/src/main/java/org/jboss/hal/testsuite/util/AssertionErrorCollector.java
rhatlapa: As discussed on our f2f meeting, this class is not needed, you can either use Rule, e.g. http://junit.org/junit4/javadoc/4.12/org/junit/rules/ErrorCollector.html
Or based on the use case, just called messages and after the cycle use `assertTrue("bla bla failed with " + failures, failures.isEmpty())`

honza-kasik:
Thanks for all suggestions! I worked them in.

zebrik:
LGTM. Thank you. Merging ...

PR #229, opened 2016-12-20
Tests for test connection in datasource wizard

honza-kasik:
* Ready for 2.9.0.Final
* Rewritten old tests

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/AbstractTestConnectionTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/AbstractTestConnectionTestCase.java
@@ -9,18 +9,27 @@ import org.jboss.hal.testsuite.fragment.config.datasource.TestConnectionWindow;
 import org.jboss.hal.testsuite.fragment.formeditor.Editor;
 import org.jboss.hal.testsuite.fragment.formeditor.PropertyEditor;
 import org.jboss.hal.testsuite.page.config.DatasourcesPage;
-import org.jboss.hal.testsuite.util.Console;
 import org.junit.Assert;
 import org.openqa.selenium.WebDriver;
-import org.wildfly.extras.creaper.core.online.operations.Address;
-import org.wildfly.extras.creaper.core.online.operations.OperationException;
-
-import java.io.IOException;
 
 /**
  * @author jcechace
  */
 public abstract class AbstractTestConnectionTestCase {
+
+    /**
+     * Hook which method will be executed after test connection was performed in wizard
+     */
+    protected interface Hook {
+        void afterConnectionWasTested(DatasourceWizard wizard);
+    }
+
+    private static final String
+            NAME = "name",
+            JNDI_NAME = "jndiName",
+            CONNECTION_URL = "connectionUrl",
+            URL = "URL",
+            DETECTED_DRIVER = "h2";
 
     @Drone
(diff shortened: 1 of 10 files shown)


honza-kasik:
@zebrik @rhatlapa Can you please take a look at this PR? The RFE is in 'Verification TODO' state.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/AbstractTestConnectionTestCase.java
```diff
@@ -9,81 +9,194 @@
...
 
 /**
  * @author jcechace
  */
 public abstract class AbstractTestConnectionTestCase {
 
+    /**
+     * Hook which method will be executed after test connection was performed in wizard
+     */
+    protected interface Hook {
```
rhatlapa: More descriptive name would be better :-)

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/AbstractTestConnectionTestCase.java
```diff
@@ -9,81 +9,194 @@
...
 
-    protected void testConnection(String name, boolean expected) {
-        datasourcesPage.view(name);
-        Console.withBrowser(browser).waitUntilLoaded();
-
+    /**
+     * Tests connection in datasource view
+     * @param expected true if connection is expected to succeed, false otherwise
+     */
+    protected void testConnection(boolean expected) {
```
rhatlapa: Better descriptive name, to make clear in what context the connection is tested

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/datasources/DomainTestConnectionTestCase.java
```diff
@@ -1,112 +1,82 @@
...
-    }
 
     @AfterClass
-    public static void tearDown() throws CommandFailedException { // remove datasources when finished
+    public static void tearDown() throws InterruptedException, TimeoutException, IOException {
         try {
-            dsOps.removeDataSource(dsNameValid);
-            new DataSourcesOperations(fullHaClient, "full-ha").removeDataSource(dsSameNameValid); ///
-            dsOps.removeDataSource(dsSameNameInvalid);
+            administration.reloadIfRequired();
```
rhatlapa: ??? teardown method which calls only reload seems as hack for something which should be already handled, but OK.

common/src/main/java/org/jboss/hal/testsuite/fragment/WindowFragment.java
```diff
@@ -87,13 +89,17 @@ public WindowState clickSave() {
...
     public void cancel() {
-        String label = PropUtils.get("modals.window.cancel.label");
-        clickButton(label);
+        clickCancel();
 
         waitUntilClosed();
         closed = true;
     }
 
+    public void clickCancel() {
```
rhatlapa: Doesn't seem as something which should be public, I would say, it should be rather private.
If it is required by cancelAndDismissReloadRequiredWindow, it should be most likely improved in order for this method not being required public

rhatlapa (commented):
Generally LGTM, just suggestions for rather minor improvements to the API.

rhatlapa:
Thanks, LGTM, merging.

PR #232, opened 2017-01-10
Reverting removing of deprecated atributes in HTTPS Listener

honza-kasik:
See https://issues.jboss.org/browse/JBEAP-7846


rhatlapa:
Thanks, LGTM, merging.

PR #234, opened 2017-01-18
Transactions subsystem metrics test

zebrik:


basic/src/test/java/org/jboss/hal/testsuite/test/runtime/transaction/TransactionsMetricsTestCase.java
```diff
@@ -0,0 +1,218 @@
...
+        JavaArchive jar = ShrinkWrap.create(JavaArchive.class, ARCHIVE_NAME + ".jar")
+                .addPackage("org.jboss.hal.testsuite.test.runtime.transaction.beans")
+                .addClass(TxnRollbackException.class)
+                .addAsManifestResource(new StringAsset("Dependencies: org.jboss.jts\n"), "MANIFEST.MF");
+        deploymentOps.deploy(jar);
+        originallyStatisticEnabled = ops.readAttribute(TXN_ADDRESS, STATISTICS_ENABLED).booleanValue();
+        if (!originallyStatisticEnabled) {
+            ops.writeAttribute(TXN_ADDRESS, STATISTICS_ENABLED, true);
+            adminOps.reloadIfRequired();
+        }
```
rhatlapa: The `originallyStatisticEnabled` should be rather remembering the `ModelNode` itself excluding default values in order to properly revert the changes after tests.

then you can always set it to true afterwards.

basic/src/test/resources/jboss-ejb-client.properties
```diff
@@ -0,0 +1,30 @@
...
+#
+remote.connectionprovider.create.options.org.xnio.Options.SSL_ENABLED=false
+
+remote.connections=jbossts
+
+remote.connection.jbossts.host=${host.url:localhost}
+remote.connection.jbossts.port=8080
+remote.connection.jbossts.connect.options.org.xnio.Options.SASL_POLICY_NOANONYMOUS=false
+remote.connection.jbossts.connect.options.org.xnio.Options.SASL_POLICY_NOPLAINTEXT=false
+
```
rhatlapa, line 30: As this is more general configuration then for transactions, I would suggest using different name for remote.connections to make it clear it is usable also for other purposes where remote ejb calls are needed. But OK.

common/src/main/java/org/jboss/hal/testsuite/util/EJBUtils.java
```diff
@@ -0,0 +1,147 @@
...
+     * @return new InitialContext.
+     * @throws NamingException
+     */
+    public static Context createNamingContext() throws NamingException {
+        final Properties jndiProps = new Properties();
+        jndiProps.setProperty(Context.URL_PKG_PREFIXES, "org.jboss.ejb.client.naming");
+        return new InitialContext(jndiProps);
+    }
+
+    public static String getNodeName() {
```
rhatlapa: Never used, don't add.

common/src/main/java/org/jboss/hal/testsuite/util/EJBUtils.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.Properties;
+import java.util.logging.Logger;
+
+import javax.naming.Context;
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
+
+public abstract class EJBUtils {
```
rhatlapa, line 31: This class is copy paste from different project, please use only parts which are relevant for you, remove the rest.

rhatlapa (changes requested):
Generally LGTM, just few rather minor improvements expected. Mainly it is not good to expose APIs, which are not needed.

zebrik:
Fixes for objections were integrated to the commit.

PR #236, opened 2017-01-23
XA datasource pool statistics test

zebrik:


zebrik:
@honza-kasik please review

honza-kasik (commented):
Thanks! Generally LGTM, although IMO deployment sources should be more separated. I suggest either:

* move all classes including `org.jboss.hal.testsuite.test.runtime.datasource.TestEntity` and classes in `org.jboss.hal.testsuite.test.runtime.datasource.beans` package to separate package or
* ~~move all deployment classes to `src/test/resources`~~

 so they cannot be confused with test sources

EDIT: I just noticed, that the second option is not possible because of import dependencies...

zebrik:
@honza-kasik requested change was pushed.

honza-kasik:
Thanks! Merging!

PR #237, opened 2017-01-26
Added API for closing wizard and dismissing reload required window

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/fragment/WindowFragment.java
```diff
@@ -70,18 +70,21 @@ public void waitUntilClosed() {
...
 
-        close.click();
+        clickCross();
 
         waitUntilClosed(windowCount, true);
 
         closed = true;
     }
 
+    protected void clickCross() {
```
rhatlapa: It seems as typo in method name, shouldn't it be `clickClose` ?

honza-kasik: This was intended since there is no "Close" button, but only the cross in top right corner. If you think it should be better to have `clickClose()`, I can fix it.

rhatlapa: I would suggest better name then `clickCross`, which would better indicate that the top the button at top right corner for closing is meant.

Thinking about possible names, I haven't figured out the best one yet, what about `clickWindowCloseButton` or something like that.

honza-kasik:
Thanks! I changed method's name!

rhatlapa:
LGTM, please squash.

honza-kasik:
Squashed!

rhatlapa:
Thanks, merging.

PR #231, opened 2017-01-05
Expanded tests for Undertow filters and reworked ResourceTable private API

honza-kasik:

PR #249, opened 2017-02-22
Fixed invalid socket binding test in JGroup*TestCase

honza-kasik:
Other failing tests in JGroups are fixed by #231


zebrik:
Thank you!

PR #259, opened 2017-03-13
Added fail message to failing tests in DatasourceNonXAPolicyTestCase

honza-kasik:


zebrik:
Merging but we should rewrite these tests to be independent since the fact that half of them now pass is wrong.

PR #260, opened 2017-03-14
JMS XA pooled connection factory metrics test

zebrik:


basic/src/test/java/org/jboss/hal/testsuite/test/runtime/messaging/XaJmsMetricsTestCase.java
```diff
@@ -0,0 +1,161 @@
...
+    private static final Logger log = LoggerFactory.getLogger(XaJmsMetricsTestCase.class);
+    private static final OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient();
+    private static final Administration adminOps = new Administration(client);
+    private static final Operations ops = new Operations(client);
+    private static final DeploymentsOperations deployOps = new DeploymentsOperations(client);
+    private static final String ARCHIVE_NAME = "xa-jms-txn", WAR_FILE_EXTENSION = ".war",
+            GENERATE_STATS_URL = "http://" + ConfigUtils.getUrl().getHost() + ":8080/" + ARCHIVE_NAME,
+            SUBSYSTEM_NAME = "messaging-activemq", CONNECTION_FACTORY_NAME = "activemq-ra", SERVER = "server",
+            DEFAULT = "default", POOLED_CONNECTION_FACTORY = "pooled-connection-factory",
+            STATISTICS_ENABLED = "statistics-enabled";
```
rhatlapa: I would suggest to put one constant per line, it is more easy to read. Just saying :-)

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/messaging/XaJmsMetricsTestCase.java
```diff
@@ -0,0 +1,161 @@
...
+    @BeforeClass
+    public static void beforeClass() throws Exception {
+        ModelNode entriesNode = new ModelNodeListBuilder(new ModelNode("java:/jms/queue/test")).build();
+        ops.add(queueAddress, Values.of("entries", entriesNode)).assertSuccess();
+        WebArchive war = ShrinkWrap.create(WebArchive.class, ARCHIVE_NAME + WAR_FILE_EXTENSION);
+        war.addPackage(MessagingStatisticsBean.class.getPackage());
+        deployOps.deploy(war);
+
+        originalFactoryStatsEnabledResult = ops.readAttribute(pooledConnectionFactoryAddress, STATISTICS_ENABLED);
+        originalFactoryStatsEnabledResult.assertSuccess();
```
rhatlapa, line 94: I believe you should call `assertDefined()` as after this line you expect it to be defined. WDYT?

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/messaging/XaJmsMetricsTestCase.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+
+    @Test
+    public void jmsXaPoolStatisticsTest() throws Exception {
+        generateStats();
+        page.navigateToDefaultProviderStats().switchToPooledConnectionFactory()
+            .selectConnectionFactory(CONNECTION_FACTORY_NAME);
+        verifyMetrics();
+        generateStats();
```
rhatlapa, line 129: What are the benefits of calling generateStats() again? The statistics were already generated on the first line of this method.

common/src/main/java/org/jboss/hal/testsuite/page/runtime/MessagingStatisticsPage.java
```diff
@@ -0,0 +1,29 @@
+package org.jboss.hal.testsuite.page.runtime;
+
+import org.jboss.hal.testsuite.page.MetricsPage;
+import org.jboss.hal.testsuite.util.Console;
+
+public class MessagingStatisticsPage extends MetricsPage {
```
rhatlapa: Javadoc :-)

rhatlapa (approved):
Please just explain the multiple callings of generateStats() and I can merge.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/messaging/XaJmsMetricsTestCase.java
zebrik: The reason is to verify functionality of the Refresh button.

zebrik: I will add comment about it to the code.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/messaging/XaJmsMetricsTestCase.java
zebrik: Yes currently there seems to always be defined value. But what if this behaviour will change in the future and there will be possible undefined value? This would be possible especially since the attribute metadata are "required" => false and "nillable" => true. 
But the following condition should be fixed accordingly. I will do it.

zebrik:
@rhatlapa please review again but don't merge. Needs to be squashed first.

rhatlapa:

zebrik:
Squashed.

rhatlapa:
Thanks, merging.

PR #261, opened 2017-03-15
Fixed BufferPoolTestCase

honza-kasik:
Default value was undefined which equals unchecked checkbox, no change was made and instead of result, exception was thrown


zebrik:
Merging.. Thx.

PR #267, opened 2017-03-22
Removed deprecated methods from ElytronIntegrationChecker

honza-kasik:
Also removed usage of depracated methods.


zebrik:
LGTM. Thx.

PR #269, opened 2017-03-22
Tweaked resource table to be easier debuggable

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/fragment/shared/table/TooFewColumnsException.java
```diff
@@ -0,0 +1,12 @@
+package org.jboss.hal.testsuite.fragment.shared.table;
+
+public class TooFewColumnsException extends RuntimeException {
+
+    public TooFewColumnsException(String message) {
```
zebrik: Do you plan to use this constructor? Please remove it if not.

zebrik (approved):
Just one optional suggestion.

common/src/main/java/org/jboss/hal/testsuite/fragment/shared/table/TooFewColumnsException.java
honza-kasik: Fixed in amended commit, thanks!

zebrik:
LGTM. Thx.

PR #271, opened 2017-03-23
Tweaked ElytronIntegrationChecker and ConfigChecker

honza-kasik:
* Added human like entering method
* Removed unnecessary code from ElytronIntegrationChecker, also now using documented public API
* Add possibility to choose input method in ConfigChecker

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/fragment/formeditor/Editor.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/fragment/formeditor/Editor.java
@@ -4,7 +4,9 @@ import org.jboss.arquillian.graphene.Graphene;
 import org.jboss.arquillian.graphene.findby.ByJQuery;
 import org.jboss.hal.testsuite.fragment.BaseFragment;
 import org.jboss.hal.testsuite.util.Console;
+import org.jboss.hal.testsuite.util.Library;
 import org.openqa.selenium.By;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/jboss/hal/testsuite/util/ConfigChecker.java
```diff
@@ -154,10 +167,20 @@ public Builder configFragment(ConfigFragment config) {
...
          */
         public Builder edit(InputType inputType, String identifier, Object attrValue) {
+            return edit(inputType, identifier, attrValue, InputMethod.MACHINE);
+        }
+
+        /**
+         * use this method only if you need to set explicit {@link InputMethod}. Set field identified by
+         * <b>{@code identifier}</b> to be edited with <b>{@code attrValue}</b>. Can be called multiple times. The
+         * actual edit will be performed as soon as client calls the {@link #andSave()} </b>
+         * @param inputMethod Input method to be used. For example InputMethod.HUMAN to edit form like human.
```
zebrik: I would suggest to extend comment to emphasise that using this method is meant just as a workaround to observed weird behaviour of some input fields and should be used only in cases when standard `edit(InputType inputType, String identifier, Object attrValue)` method is not working properly.

common/src/main/java/org/jboss/hal/testsuite/fragment/formeditor/Editor.java
```diff
@@ -51,6 +53,26 @@ public void text(String identifier, String value) {
     }
 
     /**
+     * Enters text as humanly as possible. E.g. makes pauses between characters and utilizes keyboard shortcuts.
```
zebrik: Please add note that this method should be used only as a workaround in cases when `text(String, String)` is not working properly.

zebrik (commented):
Just suggestions related to comments.

honza-kasik:
Thanks! I improved comments and this is ready to be merged now!

honza-kasik:
I need to build upon this, so I merged this myself, since errors were only in comments.

PR #273, opened 2017-03-23
Added state to WizardWindow, updated DiscoveryGroupTestCase

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/fragment/shared/modal/WizardWindow.java
```diff
@@ -159,4 +222,38 @@ public void waitUntilFinished() {
         Graphene.waitGui().withTimeout(1200, TimeUnit.MILLISECONDS);
         Graphene.waitModel().until().element(root, selector).is().not().visible();
     }
+
+    /**
+     * Providing information about state of wizard (open/closed)
+     */
+    public static final class State {
```
honza-kasik, line 224: WDYT @zebrik?

common/src/main/java/org/jboss/hal/testsuite/fragment/shared/modal/WizardWindow.java
```diff
@@ -159,4 +217,38 @@ public void waitUntilFinished() {
...
+            Assert.assertEquals("Wizard was supposed to fail, the window should be open.",
+                    InnerState.OPEN, state);
+        }
+
+    }
+
+    /**
+     * Obtains {@link State} of wizard
+     */
+    public State getWizardState() {
```
zebrik: I would suggest private modifier for this method.

zebrik (approved):
Just one minor recommendation.

common/src/main/java/org/jboss/hal/testsuite/fragment/shared/modal/WizardWindow.java
honza-kasik: Fixed, thanks!

zebrik:
LGTM. Thx.

PR #275, opened 2017-03-27
Added support for domain mode for messaging transactions tests

honza-kasik:


transactions/README.md
```diff
@@ -11,7 +9,7 @@ When running this module, the server has to have already replaced its journal by
 
 You can do this either by
 
-* running `./prepare-and-start-server.sh -t <path to prepared journal> -u <path to unzipped EAP>`
+* running `./prepare-and-start-server.sh -t <path to prepared journal> -u <path to unzipped EAP>` with optional `-d` when running tests in domain mode
```
rhatlapa: does it mean that running with `-d` means testing server in domain mode? If yes, it should be better described as currently it sounds as some option which makes sense in domain mode only without any explanation what it does.

rhatlapa:
Generally LGTM, please just update the description of the `-d` option

honza-kasik:
@rhatlapa Thanks! I fixed description and added description to script itself.

transactions/README.md
```diff
@@ -1,15 +1,18 @@
...
 
-When running this module, the server has to have already replaced its journal by the one from attached zip `artemis-journal.zip`.
+When running this module, the server has to have already replaced its journal by the one from attached zip 
+`artemis-journal.zip`.
 
 You can do this either by
 
-* running `./prepare-and-start-server.sh -t <path to prepared journal> -u <path to unzipped EAP>` with optional `-d` when running tests in domain mode
+* running `./prepare-and-start-server.sh -t <path to prepared journal> -u <path to unzipped EAP>` which also starts and
+unsecures server. Use optional `-d` to prepare and start server in domain mode instead of standalone.
```
zebrik: I would suggest to reword from  

> Use optional `-d` to prepare 

to something like 

> Use optional `-d` argument to prepare

zebrik (commented):
Just one minor suggestion. I don't insist on it. If you squash I will merge.

honza-kasik:
@zebrik Thanks! I updated description and squashed commits!

zebrik:
Thank you! Merging..

PR #277, opened 2017-04-04
 	Tests for Elytron transformer configuration

zebrik:


honza-kasik (approved):
LGTM, just a very minor note: Do you think that it is best solution just to fail test with 'Assert.fail()'?

I can merge it right away...

zebrik:
`Assert.fail()` is nasty enough to force me to fix it as soon as possible ;-)

honza-kasik:
Ok, merging :)

honza-kasik:
Thanks!

PR #281, opened 2017-04-13
Fixed mail server tests

honza-kasik:
* Added test for tls attribute
* Fixed socket binding test
* Reported issue https://issues.jboss.org/browse/HAL-1349


rhatlapa:
LGTM, thanks, merging.

PR #282, opened 2017-04-13
Added tests for Elytron security realms

honza-kasik:
* Code is ready
* Test documentation is missing for now, I will add it if it is required

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/securityrealm/ElytronJDBCSecurityRealmTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/securityrealm/ElytronJDBCSecurityRealmTestCase.java
@@ -32,6 +32,13 @@ public class ElytronJDBCSecurityRealmTestCase extends AbstractElytronTestCase {
     @Page
     private SecurityRealmPage page;
 
+    /**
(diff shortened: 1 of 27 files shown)


honza-kasik:
Added documentations to test cases

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/securityrealm/ElytronJDBCSecurityRealmTestCase.java
```diff
@@ -0,0 +1,266 @@
...
+
+        public ModelNode getPrincipalQueries() {
+            return listBuilder.build();
+        }
+
+        public Address getAddress() {
+            return address;
+        }
+
+        public static final class Builder {
```
zebrik: I don't suggest to use `public` visibility for class (Builder) inside of a `private` class (JDBCRealm)  ;-) 
Moreover both of them are and should be called only from the top level class (ElytronJDBCSecurityRealmTestCase) so there is no reason in making any of them public.

zebrik (commented):
LGTM except of one minor complaint about nested class visibility.

honza-kasik:
Thanks! I made Builder `private`.

zebrik:
LGTM. @honza-kasik do you feel it's suitable to do (partial ?) squashing in this case?

PR #283, opened 2017-04-18
EAP7-583: Webconsole show content of archive even if deployment is not exploded

jstefl:
Instead of failing if the read-content operation is invoked for archived content or with a path parameter indicating a path inside archived child content, instead extract the content from the archive and return it.


rhatlapa:
@zebrik The CI seems to failed due creaper downloading, I shall take a look => assigning myself for review.
Otherwise the changes LGTM.

zebrik:
@rhatlapa seems to be just temporary repository.jboss.org issue. I've seen this before.

rhatlapa:
@zebrik you are right just temporary outage, now passed, merging.

PR #287, opened 2017-04-20
Checkstyle update

honza-kasik:
* Generates XML report
* Generates HTML site
* Also output for console
* Configuration made according to documentation
* Works for me!


honza-kasik:
@zebrik WDYT?

honza-kasik:
Just one note - we should be running only 'check' goal during 'verify' or 'validate' phase, but since it just outputs to console I used 'checkstyle'/'checkstyle-aggregate' which generates XML and HTML report instead in 'compile' phase.

zebrik:
Looks OK. Please remove the 'Dummy fail' commit.

honza-kasik:
Thanks, removed!

zebrik:
Thank you!

PR #290, opened 2017-04-27
Added elytron authentication tests

honza-kasik:
* Upgraded Creaper version - now with configuration backups!
* Added method to ResourceVerifier to verify if list attribute contains item

### DIFF
--- a/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/DefaultAuthenticationContextTestCase.java
+++ b/basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/DefaultAuthenticationContextTestCase.java
@@ -1,0 +1,104 @@
+package org.jboss.hal.testsuite.test.configuration.elytron.authentication;
+
+import org.apache.commons.lang.RandomStringUtils;
+import org.jboss.arquillian.graphene.page.Page;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.hal.testsuite.page.config.elytron.ElytronAuthenticationPage;
+import org.jboss.hal.testsuite.test.configuration.elytron.AbstractElytronTestCase;
+import org.jboss.hal.testsuite.test.configuration.elytron.ElytronOperations;
+import org.jboss.hal.testsuite.util.ConfigChecker;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.extras.creaper.commands.foundation.online.SnapshotBackup;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+
+import java.io.IOException;
+
+@RunWith(Arquillian.class)
+public class DefaultAuthenticationContextTestCase extends AbstractElytronTestCase {
+
+    @Page
+    private ElytronAuthenticationPage page;
+
+    private static final String
+        AUTHENTICATION_CONTEXT = "authentication-context",
+        DEFAULT_AUTHENTICATION_CONTEXT = "default-authentication-context",
+        FINAL_PROVIDERS = "final-providers",
+        INITIAL_PROVIDERS = "initial-providers";
+
+    /**
+     * @tpTestDetails Try to edit default-authentication-context attribute value in Web Console's Elytron subsystem
+     * configuration.
+     * Validate edited attribute value in the model.
+     */
+    @Test
+    public void editDefaultAuthenticationContext() throws Exception {
+        final Address authenticationContextAddress = createAuthenticationContext();
+
+        final SnapshotBackup snapshotBackup = new SnapshotBackup();
+        client.apply(snapshotBackup.backup());
+        try {
+            page.navigate();
+            new ConfigChecker.Builder(client, ElytronOperations.getElytronSubsystemAddress())
+                    .configFragment(page.getConfigFragment())
+                    .editAndSave(ConfigChecker.InputType.TEXT, DEFAULT_AUTHENTICATION_CONTEXT, authenticationContextAddress.getLastPairValue())
+                    .verifyFormSaved()
+                    .verifyAttribute(DEFAULT_AUTHENTICATION_CONTEXT, authenticationContextAddress.getLastPairValue());
+        } finally {
+            client.apply(snapshotBackup.restore());
+        }
+    }
+
+    /**
+     * @tpTestDetails Try to edit final-providers attribute value in Web Console's Elytron subsystem configuration.
+     * Validate edited attribute value in the model.
+     */
+    @Test
+    public void editFinalProviders() throws Exception {
+        final String providerLoader = RandomStringUtils.randomAlphanumeric(7);
+
+        final SnapshotBackup snapshotBackup = new SnapshotBackup();
+        client.apply(snapshotBackup.backup());
+        try {
+            elyOps.addProviderLoader(providerLoader);
+            page.navigate();
+            new ConfigChecker.Builder(client, ElytronOperations.getElytronSubsystemAddress())
+                    .configFragment(page.getConfigFragment())
+                    .editAndSave(ConfigChecker.InputType.TEXT, FINAL_PROVIDERS, providerLoader)
+                    .verifyFormSaved()
+                    .verifyAttribute(FINAL_PROVIDERS, providerLoader);
+        } finally {
+            client.apply(snapshotBackup.restore());
+        }
+    }
+
+    /**
+     * @tpTestDetails Try to edit initial-providers attribute value in Web Console's Elytron subsystem configuration.
+     * Validate edited attribute value in the model.
+     */
+    @Test
+    public void editInitialProviders() throws Exception {
+        final String providerLoader = RandomStringUtils.randomAlphanumeric(7);
+
+        final SnapshotBackup snapshotBackup = new SnapshotBackup();
+        client.apply(snapshotBackup.backup());
+        try {
+            elyOps.addProviderLoader(providerLoader);
+            page.navigate();
+            new ConfigChecker.Builder(client, ElytronOperations.getElytronSubsystemAddress())
+                    .configFragment(page.getConfigFragment())
+                    .editAndSave(ConfigChecker.InputType.TEXT, INITIAL_PROVIDERS, providerLoader)
+                    .verifyFormSaved()
+                    .verifyAttribute(INITIAL_PROVIDERS, providerLoader);
+        } finally {
+            client.apply(snapshotBackup.restore());
+        }
+    }
+
+    private Address createAuthenticationContext() throws IOException {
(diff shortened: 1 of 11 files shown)


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/DefaultAuthenticationContextTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+     * @tpTestDetails Try to edit default-authentication-context attribute value in Web Console's Elytron subsystem
+     * configuration.
+     * Validate edited attribute value in the model.
+     */
+    @Test
+    public void editDefaultAuthenticationContext() throws Exception {
+        final Address authenticationContextAddress = createAuthenticationContext();
+
+        final SnapshotBackup snapshotBackup = new SnapshotBackup();
+        client.apply(snapshotBackup.backup());
```
zebrik: What about moving snapshot backup/restore logic to Before/After methods? Since we are developing integration tests I don't believe that we ever want to run our tests in parallel so IMHO we can keep some state (model snapshot) outside of test methods. @rhatlapa WDYT?

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/DefaultAuthenticationContextTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+        INITIAL_PROVIDERS = "initial-providers";
+
+    /**
+     * @tpTestDetails Try to edit default-authentication-context attribute value in Web Console's Elytron subsystem
+     * configuration.
+     * Validate edited attribute value in the model.
+     */
+    @Test
+    public void editDefaultAuthenticationContext() throws Exception {
+        final Address authenticationContextAddress = createAuthenticationContext();
```
zebrik, line 37: Why is this called before backup?

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/ElytronAuthenticationConfigurationTestCase.java
```diff
@@ -0,0 +1,528 @@
...
+     * @tpTestDetails Create Elytron Authentication configuration instance in model and try to edit its
+     * allow-all-mechanisms attribute value in Web Console's Elytron subsystem configuration.
+     * Validate edited attribute value in the model.
+     */
+    @Test
+    public void toggleAllowAllMechanisms() throws Exception {
+        final SnapshotBackup snapshotBackup = new SnapshotBackup();
+        client.apply(snapshotBackup.backup());
+
+        final Address authenticationConfigurationAddress = createAuthenticationConfiguration();
```
zebrik: Why is this called outside of the try block?
This questions applies as well for many further parts of code.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/ElytronAuthenticationConfigurationTestCase.java
```diff
@@ -0,0 +1,528 @@
...
+                    .addResource(AddAuthenticationConfigurationWizard.class)
+                    .name(authenticationConfigurationAddress.getLastPairValue())
+                    .saveAndDismissReloadRequiredWindowWithState()
+                    .assertWindowClosed();
+
+            Assert.assertTrue(page.getResourceManager().isResourcePresent(authenticationConfigurationAddress.getLastPairValue()));
+
+            new ResourceVerifier(authenticationConfigurationAddress, client).verifyExists();
+        } finally {
+            ops.removeIfExists(authenticationConfigurationAddress);
```
zebrik, line 75: Just curious ... how do you decide whether to use snapshots or direct creaper cleanup operations?

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -236,6 +236,44 @@ public ResourceVerifier verifyAttributeIsUndefined(String attributeName) throws
...
+            return actualResult.isSuccess() && actualResult.hasDefined(Constants.RESULT);
+        });
+
+        final ModelNodeResult modelNodeResult = ops.readAttribute(resourceAddress, attributeName);
+        modelNodeResult.assertSuccess();
+
+        Assert.assertTrue("Given value '" + value.toString() + "' is not present in list attribute '" + attributeName + "'!" +
+                        (errorMessageSuffix == null || errorMessageSuffix.isEmpty() ? "" : " " + errorMessageSuffix),
+                modelNodeResult.listValue().stream()
+                        .peek(modelNode -> log.debug("Comparing '{}' with list member '{}'.", modelNode.toString(), value.toString()))
```
zebrik: How many data does this log? What about trace log level instead? Just asking...

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -236,6 +236,44 @@ public ResourceVerifier verifyAttributeIsUndefined(String attributeName) throws
...
     }
 
+    /**
+     * Verifies that list type attribute contains give value.
+     * @param value Value which should be present in the list.
+     */
+    public ResourceVerifier verifyListAttributeContainsValue(String attributeName, ModelNode value, String errorMessageSuffix) throws Exception {
+        waitFor(() -> {
+            ModelNodeResult actualResult = ops.readAttribute(resourceAddress, attributeName);
+            return actualResult.isSuccess() && actualResult.hasDefined(Constants.RESULT);
```
zebrik: Is it sufficient to just verify that the attribute is readable and not that expected list item is already propagated to it?

zebrik (commented):
Several questions. Nothing serious. See inline.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/DefaultAuthenticationContextTestCase.java
honza-kasik: That's definitely better! We can call backup once in `BeforeClass` and then restore in `After`.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/ElytronAuthenticationConfigurationTestCase.java
honza-kasik: I started to use backup operations after I found out Creaper 1.6.0 is out :)

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/ElytronAuthenticationConfigurationTestCase.java
honza-kasik: Because if it fails, there is no resource added - thus nothing to remove in `finally`

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
honza-kasik: There are usually under five records in table, so it is bearable.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/ElytronAuthenticationConfigurationTestCase.java
```diff
@@ -180,32 +185,33 @@ public void editAllowSASLMechanisms() throws Exception {
...
+                                .verifyFormSaved()
+                                .verifyAttribute(ANONYMOUS, !initialValue);
+
+                        Console.withBrowser(browser).dismissReloadRequiredWindowIfPresent();
+
+                        new ConfigChecker.Builder(client, authenticationConfigurationAddress)
+                                .configFragment(page.getConfigFragment())
+                                .editAndSave(ConfigChecker.InputType.CHECKBOX, ANONYMOUS, initialValue)
+                                .verifyFormSaved()
+                                .verifyAttribute(ANONYMOUS, initialValue);                    },
```
zebrik: please remove the spaces before `}`. Ideally add new line before the `}` instead of the spaces.

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/authentication/ElytronAuthenticationOperations.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * Back ups values of attributes, performs and {@link Action} and reverts the attributes to their original value.
+     *
+     * @param address        Address of attributes
+     * @param action         Action performed with backed up attributes
+     * @param attributes     Names of attributes
+     */
+    public void performActionOnAttributeAndRevertItsValueToOriginal(Address address, Action action, String... attributes) throws Exception {
```
zebrik: Do you really plan to use attributes as String array? Since in the PR you always pass just one String. I'm asking since if you use `String attribute` parameter instead of `String... attributes` array parameter you can switch the order of the last 2 parameters (in that case parameters would be `Address address, String attribute, Action action`) and the usage of this method would be IMHO much more readable without attribute name lost somewhere after lambda body ... just suggestion.

zebrik (commented):
Just 2 suggestions. I let it to you to decide if you fix it. Then please squash and I'll merge.

common/src/main/java/org/jboss/hal/testsuite/creaper/ResourceVerifier.java
```diff
@@ -236,6 +236,85 @@ public ResourceVerifier verifyAttributeIsUndefined(String attributeName) throws
...
+    /**
+     * Verifies that list type attribute contains give value.
+     * @param value Value which should be present in the list.
+     */
+    public ResourceVerifier verifyListAttributeDoesNotContainValue(String attributeName, ModelNode value, String errorMessageSuffix)
+            throws Exception {
+        waitFor(() -> {
+            ModelNodeResult actualResult = ops.readAttribute(resourceAddress, attributeName);
+            return actualResult.isSuccess() &&
+                    actualResult.hasDefined(Constants.RESULT);
```
zebrik: I suggest to add
`&& !isModelNodePresentInListAttributeValue(attributeName, value)`
to the condition.

zebrik (commented):
One very last suggestion:

honza-kasik:
Thanks! I expanded condition. Can I squash it now?

zebrik:
@honza-kasik yes, please squash it.

honza-kasik:
Squashed!

zebrik:
Thank you!

PR #291, opened 2017-05-03
WFCORE-2427: Elytron credential-store resource was refactored.

zebrik:
THe `uri` super-attribute was finally removed from `credential-store` and replaced with several new attributes. GUI changed accordingly. 
Moreover only lower letters seems to be allowed in the alias name.


zebrik:
@honza-kasik - just FYI.

PR #299, opened 2017-05-31
Added tests for identity mappings in LDAP security realm

honza-kasik:


common/src/main/java/org/jboss/hal/testsuite/creaper/ModelNodeUtils.java
```diff
@@ -0,0 +1,27 @@
+package org.jboss.hal.testsuite.creaper;
```
zebrik: I would suggest to move this class to the `org.jboss.hal.testsuite.dmr` package (there is already e.g. `ModelNodeGenerator` in this package)

zebrik (approved):
There is only one minor suggestion from my side.

honza-kasik:
Thanks! I moved the class to `dmr` package

PR #303, opened 2017-06-05
Fixed failing tests in JournalTestCase

honza-kasik:
* Added option for wizard window to ConfigChecker in case the ConfigFragment is present inside one.


zebrik:
LGTM. Thank you!

PR #328, opened 2017-07-03
Making page load timeout configurable.

zebrik:


common/src/main/java/org/jboss/hal/testsuite/finder/FinderNavigation.java
```diff
@@ -106,6 +109,17 @@ public FinderNavigation(final WebDriver browser, final Class<? extends BasePage>
     }
 
     /**
+     * Sets the timeout in seconds for the application to be loaded in the beginning of the navigation after refresh.
+     * Use carefully in exceptional situations where default timeout of 30 seconds is not sufficient,
+     * e.g. for large domain performance tests.
+     * @param pageLoadTimeout
```
rhatlapa: Does it have any benefit to have here the `@param pageLoadTimeout` without description? => Give there short description that it represents timeout in seconds even though you can guess it from general description.

common/src/main/java/org/jboss/hal/testsuite/util/Console.java
```diff
@@ -34,15 +34,30 @@
...
         console.browser = browser;
 
         return console;
     }
 
+    /**
+     * Sets the timeout in seconds for the application to be loaded.
+     * Use carefully in exceptional situations where default timeout of 30 seconds is not sufficient,
+     * e.g. for large domain performance tests.
+     * @param pageLoadTimeout
```
rhatlapa: The same as above.

rhatlapa:
LGTM, just minor improvement in javadoc requested. please update, and squash.

zebrik:
Parameters described in javadoc and change amended.

PR #336, opened 2017-07-19
WFCORE-3017: It's not possible to call Elytron credential-store :add-…

zebrik:
…alias via profile in domain any more.

### DIFF
--- a/common/src/main/java/org/jboss/hal/testsuite/util/ElytronIntegrationChecker.java
+++ b/common/src/main/java/org/jboss/hal/testsuite/util/ElytronIntegrationChecker.java
@@ -1,5 +1,7 @@
 package org.jboss.hal.testsuite.util;
 
+import static org.wildfly.extras.creaper.core.online.Constants.SUBSYSTEM;
+
 import org.apache.commons.lang3.RandomStringUtils;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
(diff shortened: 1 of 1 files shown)


zebrik:
@honza-kasik any objection?

honza-kasik:
LGTM, thanks!

common/src/main/java/org/jboss/hal/testsuite/util/ElytronIntegrationChecker.java
```diff
@@ -190,6 +194,46 @@ private Address addCredentialStore(String credentialStoreName) throws IOExceptio
...
+        }
+    }
+
+    private List<String> allRunningServersOnDefaultHost() throws IOException {
+        Address hostAddress = Address.host(ConfigUtils.getDefaultHost());
+        ModelNodeResult result = operations.readChildrenNames(hostAddress, Constants.SERVER);
+        result.assertDefinedValue();
+        List<String> servers = result.stringListValue();
+
+        List<String> startedServers = new ArrayList<String>();
```
rhatlapa: Creaper has methods like org.wildfly.extras.creaper.core.online.operations.admin.DomainAdministration#allRunningServers(java.lang.String), why don't you make use of it?

rhatlapa (changes requested):

rhatlapa:
It is better to reuse and not to write new one => please reuse existing code from creaper instead of writing new methods.

common/src/main/java/org/jboss/hal/testsuite/util/ElytronIntegrationChecker.java
zebrik: I wasn't aware of that method. Thx.

zebrik:
Amended.

PR #340, opened 2017-08-03
Enabled Elytron Policy TestCase and added tests

spriadka:


honza-kasik (changes requested):
Please add documentation for Elytron test methods, also, is the restart in `createPolicy(Address address)` really necessary? @zebrik WDYT?

honza-kasik (approved):

PR #346, opened 2017-08-18
Hal 1374

spriadka:


basic/src/test/java/org/jboss/hal/testsuite/test/runtime/server/ForceShutdownOperationTestCase.java
```diff
@@ -0,0 +1,195 @@
...
+            Console.withBrowser(browser).openedWindow(ConfirmationWindow.class).confirm();
+            Console.withBrowser(browser).waitUntilLoaded();
+            new ResourceVerifier(serverAddress, client).verifyAttribute(SERVER_STATE, "STOPPED");
+        } finally {
+            domainAdministration.reloadIfRequired();
+        }
+    }
+
+    @Test
+    public void testForceShutdownInLoad() throws Exception {
```
rhatlapa: Provide test case documentation similarly as is done e.g. for org.jboss.hal.testsuite.test.runtime.messaging.XaJmsMetricsTestCase#jmsXaPoolStatisticsTest

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/server/ForceShutdownOperationTestCase.java
```diff
@@ -0,0 +1,195 @@
...
+    private void loadServer() {
+        final String serverUrl = String.format("http://127.0.0.1:%d/%s%s", SERVER_PORT,
+                DEPLOYMENT_NAME, InfiniteServlet.URL_PATTERN);
+        try (CloseableHttpClient client = HttpClientBuilder.create().build()) {
+            while (!Thread.currentThread().isInterrupted()) {
+                Thread.sleep(100);
+                HttpGet httpGet = new HttpGet(serverUrl);
+                CloseableHttpResponse response = client.execute(httpGet);
+                HttpEntity entity = response.getEntity();
+                LOGGER.debug(EntityUtils.toString(entity));
```
rhatlapa: Wouldn't it be more beneficial to just check/log response code? Instead of whole response? Also if you are reading it the entity, you don't need to consumeQuietly the entity again as you've already read it.

basic/src/test/java/org/jboss/hal/testsuite/test/runtime/server/InfiniteServlet.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+@WebServlet("/infinity")
+public class InfiniteServlet extends HttpServlet {
```
rhatlapa: Personally I would prefer the name of servlet to match what the servlet does and not naming it just based on test purpose.

From its name I would expect that it will do something for infinity, which doesn't correspond to reality.

common/src/main/java/org/jboss/hal/testsuite/creaper/command/UndeployCommand.java
```diff
@@ -48,7 +48,7 @@ public void apply(OnlineCommandContext ctx) throws Exception {
             if (fromAllGroups) {
                 cmd.append(" --all-relevant-server-groups");
             } else {
-                cmd.append("--server-groups").append(particularGroup);
+                cmd.append(" --server-groups= ").append(particularGroup);
```
rhatlapa: this change doesn't seem to be fully correct, I believe there shouldn't be space after the `=`. WDYT?

rhatlapa (changes requested):

spriadka:
@rhatlapa fixed requested changes. Can you have another look into it please?

rhatlapa (approved):

PR #349, opened 2017-09-15
Hal 1206

spriadka:


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/general/WebAccessTestCase.java
```diff
@@ -26,15 +23,18 @@
  * @author mkrajcov <mkrajcov@redhat.com>
  */
 @RunWith(Arquillian.class)
-@Category(Shared.class)
+@RunAsClient
```
zebrik, line 26: Why is `RunAsClient` annotation needed here?

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/JCA/DatasourceNonXAPolicyTestCase.java
```diff
@@ -41,13 +34,12 @@
     private static final OnlineManagementClient client = ManagementClientProvider.createOnlineManagementClient();
```
zebrik, line 37: I don't see any cleaning up after particular test methods in `DatasourceNonXAPolicyTestCase`. The model changes seem to stay persisted. Am I missing something?

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/container/IIOPTestCase.java
```diff
@@ -68,15 +71,14 @@
  * @author Harald Pehl
  */
 @RunWith(Arquillian.class)
+@RunAsClient
```
zebrik, line 74: Why `RunAsClient`?

zebrik (changes requested):

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/general/WebAccessTestCase.java
rhatlapa: Arquillian by default is running in container. If the graphene extension isn't changing it somehow, I believe we should have it for all classes defined as `@RunAsClient`. => it is most likely correct change from my PoV.

spriadka: As I understand testing with Arquillian, you should use @RunAsClient when you want to specify, that the test is not deploying any archive to the container. (https://docs.jboss.org/arquillian/spi/1.0.3.Final/org/jboss/arquillian/container/test/api/RunAsClient.html). Speaking of functionality it does not change a thing

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/JCA/DatasourceNonXAPolicyTestCase.java
spriadka: It was my mistake, fixed with 752417f

basic/src/test/java/org/jboss/hal/testsuite/test/configuration/general/WebAccessTestCase.java
zebrik: As long as we don't use `@Deployment` it's not important IMO. From my POV it's just one more line of code which is not necessary. We should either add the annotation to every test or avoid using it to prevent confusion.

zebrik: I'm just complaining about adding the annotation as part of the commit. By doing it this way you could make an impression that it is somehow needed for the commit logic.

zebrik (approved):

zebrik:
Since the cleanup after `DatasourceNonXAPolicyTestCase` is fixed this can be squashed a bit and merged. Regarding the `RunAsClient` thing I don't care much so decide as you wish.

spriadka:
Squashing and merging

PR #354, opened 2017-09-25
HTTPSListenerTestCase fixes + refactoring

spriadka:


zebrik (commented):
The `AbstractAddHTTPListenerWizard` stuff may be a little bit confusing but I'm ok with that. 
But commit message `removed @Before methods` could be more specific IMO - at least add something like `... from HTTP(S) listener test cases` 
You can, of course, skip this comment if you want to squash the commmits before merging.

PR #356, opened 2017-10-18
Elytron factory test case fixes

spriadka:


common/src/main/java/org/jboss/hal/testsuite/util/WizardValidator.java
```diff
@@ -0,0 +1,34 @@
...
+import org.jboss.hal.testsuite.fragment.shared.modal.WizardWindow;
+import org.jboss.hal.testsuite.util.configurationchanges.WizardInput;
+import org.junit.Assert;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class WizardValidator<T extends WizardWindow> {
```
zebrik: Please add Javadoc.

common/src/main/java/org/jboss/hal/testsuite/util/WizardValidator.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private Set<WizardInput<T>> inputs;
+    private ConfigFragment windowFragment;
+
+    public WizardValidator(Collection<WizardInput<T>> inputs, ConfigFragment fragment) {
```
zebrik: Please add Javadoc.

common/src/main/java/org/jboss/hal/testsuite/util/WizardValidator.java
```diff
@@ -0,0 +1,34 @@
...
+        this.inputs = new HashSet<>(inputs);
+        this.windowFragment = fragment;
+    }
+
+    public void testInputCombinationAndAssert(Collection<WizardInput<T>> presentInputs, T wizard) {
```
zebrik: Please add Javadoc.

zebrik: This method should iterate over all mandatory fields, try to fill only the particular field and verify that the rest of mandatory fields show a warning.

zebrik (changes requested):

common/src/main/java/org/jboss/hal/testsuite/util/WizardValidator.java
zebrik: I would suggest renaming of the `inputs` parameter to something like `mandatoryInputs`.

common/src/main/java/org/jboss/hal/testsuite/util/WizardValidator.java
```diff
@@ -11,20 +11,36 @@
...
+        this.mandatoryInputs = new HashSet<>(mandatoryInputs);
         this.windowFragment = fragment;
     }
 
-    public void testInputCombinationAndAssert(Collection<WizardInput<T>> presentInputs, T wizard) {
+    /**
+     * Tests all possible invalid combinations for required fields in wizard
+     * @param wizard to be tested
+     */
+    public abstract void testInvalidCombinationsAndAssert(T wizard);
```
zebrik, line 33: This is a bit more laborious than my original idea but it will work as well. We could add default implementation of this method in the future if needed.

zebrik (approved):
LGTM. Please squash.

PR #362, opened 2017-11-09
Transformer test case refactoring

spriadka:


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/transformer/PrincipalTransformerUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.IOException;
+
+import static org.apache.commons.lang3.RandomStringUtils.randomAlphanumeric;
+
+public class PrincipalTransformerUtils {
+
+    private static final String CONSTANT = "constant";
+
+    private final Administration administration;
```
zebrik: Why is the `administration` field declared? It seems not to be used.

zebrik (approved):
There is only one tiny complaint about a useless private field to be optionally fixed. Then please squash and merge.

PR #363, opened 2017-11-09
fixed HttpAuthenticationTestCase and SaslAuthenticationTestCase, remo…

spriadka:
…ved ElytronAuthenticationTestCase


basic/src/test/java/org/jboss/hal/testsuite/test/configuration/elytron/ElytronOperations.java
```diff
@@ -90,6 +90,13 @@ public Address createSecurityDomain(String realmName) throws IOException {
         return domainAddress;
     }
 
+    public void createSecurityDomain(Address securityDomainAddress) throws IOException {
```
zebrik: What about passing just `String securityDomainName` parameter and returning security domain address - and encapsulate the previous IMHO a bit useless client call of `elyOps.getElytronAddress(SECURITY_DOMAIN, securityDomainName)` ?

zebrik (commented):
There is only one small complaint to be optionally fixed. Then please squash and merge.

PR #365, opened 2017-12-06
added KnownIssue category, annotated known issues

spriadka:


pom.xml
```diff
@@ -354,6 +354,9 @@
         <test.included.category>
           org.jboss.hal.testsuite.category.Elytron
         </test.included.category>
+        <test.excluded.category>
```
zebrik: We need to be able to choose if we want to run KnownIssue tests or not. This way of implementation excludes KnownIssues tests every time. 
We rather need new profile which let us choose if we want to run them or not.

zebrik (changes requested):

pom.xml
```diff
@@ -367,24 +364,29 @@
...
           org.jboss.hal.testsuite.category.PicketBox,
           org.jboss.hal.testsuite.category.PicketLink
         </test.excluded.category>
       </properties>
     </profile>
 
+    <profile>
+      <id>withoutKnownIssues</id>
+      <properties>
+        <test.excluded.category>
```
zebrik, line 384: There seems to be conflict of `test.excluded.category` property between `noSecurity` and `withoutKnownIssues` profiles. If you use them both it would be IMO overwritten.

zebrik (changes requested):

PR #374, opened 2018-06-13
[HAL-1473]  - In Management Console - Jboss domain controller deployment units under tab ServerGroup are showing as server not running even though they are.

adamecp:
JIRA issue [https://issues.jboss.org/browse/HAL-1473](https://issues.jboss.org/browse/HAL-1473)

Description : 

In Domain mode under Management Console - Jboss domain controller deployment units under tab ServerGroup are showing as server not running even though they are.

In domain controller admin console, select Deployments; Server Groups; and Deployments column all ears show red - server not running. But they are running.

This does not affect the functionality of the server. The customer is able to successfully access the deployed applications.

Note: The issue seems intermittent as it is reproducible on one system locally and is consistent but not on other one. Both system had the fresh EAP 7.1.1 installation. However, it is consistent to the customer.

Please let me know if any additional information is required.


zebrik:
LGTM
Thank you.

